PR #10, opened 2020-02-17
Remove timestamp from date printed by CLI

mborjiginwang-Ocient:
DB-10045

### DIFF
--- a/src/main/java/com/ocient/cli/CLI.java
+++ b/src/main/java/com/ocient/cli/CLI.java
@@ -1942,7 +1942,7 @@ public class CLI {
 						String timeStr = sdf.format(o);
 						o = timeStr;
 					} else if (o instanceof Date) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1942,7 +1942,7 @@ private static void printResultSet(final ResultSet rs, final ResultSetMetaData m
 						String timeStr = sdf.format(o);
 						o = timeStr;
 					} else if (o instanceof Date) {
-						SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS");
+						SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
```
jasonar81: This won't work.  Try a query that has a timestamp and a date in it.  I believe your change will cause both timestamps and date columns to only show the date piece.

PR #16, opened 2020-02-25
JDBC tickets

mborjiginwang-Ocient:
DB-10119
DB-10120
DB-10137

### DIFF
--- a/src/main/java/com/ocient/jdbc/XGStatement.java
+++ b/src/main/java/com/ocient/jdbc/XGStatement.java
@@ -44,14 +44,15 @@ import java.util.TimeZone;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 import com.google.common.base.Preconditions;
 import com.google.protobuf.util.JsonFormat;
 
 public class XGStatement implements Statement
 {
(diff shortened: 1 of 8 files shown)


src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -280,9 +279,31 @@ public void addBatch(final String sql) throws SQLException {
...
+		// See startTask
+        synchronized (conn) {
+            conn.setQueryCancelled(true);
+            if (conn == null || conn.getRunningQueryThread() == null || conn.getRunningQueryId()== null) {
+                return;
+            }
+        }
+		try {
+			conn.getRunningQueryThread().interrupt();
+
```
leogreca: I don't think we need to call interrupt(). 

the reconnect() method closes the socket and that should cause the thread that is calling send in.read() to throw an IOException

src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -89,7 +90,26 @@ private static int bytesToInt(final byte[] val) {
...
+	public void setQueryCancelled(boolean b) { queryCancelled.set(b); }
+    public void setRunningQueryThread(Thread t) { runningQueryThread.set(t); }
+	public void passUpCancel(boolean clearCancelFlag) throws SQLException
+	{
+		synchronized (this)
+		{
+			boolean cancelled = queryCancelled.get();
+			if (clearCancelFlag)
+			{
+				cancelled = false;
```
leogreca: queryCancelled.set(false);

PR #20, opened 2020-03-10
Updated JDBC to support new DECIMAL string format

mra-ocient:


leogreca:
Please add an entry to our release notes. 
https://github.com/Xeograph/ocient_jdbc/blob/master/userdoc/release_notes.adoc

PR #41, opened 2020-07-06
WIP Build dependencies in autodeploy gitlab job

esaupe-Ocient:
This does not work on centos-build7 and 8, but works on 3 and 4.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,5 +1,7 @@
 build-jar:
   script:
-    echo hello world
+    -echo hello world
+    -pwd
+    -ping i01-fil-001 -c 2
 #  rules:
 #    - if: '$CI_COMMIT_BRANCH == "master"'
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,5 @@
+build-jar:
+  script:
+    echo hello world
+#  rules:
+#    - if: '$CI_COMMIT_BRANCH == "master"'
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,4 +1,10 @@
-build-jar:
+stages:
+  - build-dependencies
+  - build-jar
+  - clean
+
+build-deps:
+  stage: build-dependencies
   tags:
     #it likes to run on a shared runner without this and you lose network access
     - lightweight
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -0,0 +1,60 @@
...
+    - which unzip
+
+    #set versions
+    - PROTOC_VERSION=3.5.1
+    - JDK_VERSION=13.0.1
+    - MVN_VERSION=3.3.9
+
+    # build protobuf
+    - wget -nv http://i01-fil-001/xgsrc/ext/protobuf-$PROTOC_VERSION.tar.gz
+    - tar xzvf protobuf-${PROTOC_VERSION}.tar.gz > /dev/null
```
bbatliner-ocient, line 31:
```suggestion
    - tar xzf protobuf-${PROTOC_VERSION}.tar.gz
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,60 @@
...
+    - export PATH=$PATH:"${JAVA_HOME}/bin"
+    # download maven
+    - wget -nv https://archive.apache.org/dist/maven/maven-3/${MVN_VERSION}/binaries/apache-maven-${MVN_VERSION}-bin.tar.gz
+    - tar xzvf apache-maven-${MVN_VERSION}-bin.tar.gz > /dev/null
+    - rm apache-maven-${MVN_VERSION}-bin.tar.gz
+    - export PATH=$PATH:"$(pwd)/apache-maven-${MVN_VERSION}/bin"
+
+    - protoc --version
+    - javac -version
+    - mvn --version
```
bbatliner-ocient, line 54: I'm somewhat curious why the build system for the project can't handle installing protoc, javac, and maven? Is it because the build system _is_ Maven? So it implicitly needs Java and Maven just to run at all?

.gitlab-ci.yml
```diff
@@ -0,0 +1,60 @@
...
+
+    - mvn package
```
bbatliner-ocient, line 56: Does this deploy the jar anywhere? I don't know how Maven works.

esaupe-Ocient: This just builds the jar and nothing else. Everything in this PR is just testing if I have all of the build dependencies.

.gitlab-ci.yml
esaupe-Ocient: Maven is the buildsystem, so it the CI task definitely needs to download binaries for maven to accomplish anything.  I think maven implicitly needs at least a JRE to execute its build tasks, so it doesn't seem reasonable to use maven to download a jdk. 
With some effort I could probably use maven to build protoc.

PR #49, opened 2020-07-22
DB-12113 Automated jar signing

esaupe-Ocient:


.gitlab-ci.yml
```diff
@@ -6,6 +6,8 @@ build-deps:
   tags:
     #gitlab runs on a shared runner without this and you lose network access
     - lightweight
+    #force centos-build2
```
bbatliner-ocient, line 9: No need to be specific here with the comment. The tag is "declarative" - the job right now happens to run on centos-build2 because it satisfies the tag, but it could run anywhere.

PR #53, opened 2020-08-14
Add TLS support

daveboutcher:


src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -72,6 +77,13 @@
 public class XGConnection implements Connection {
 	private static final Logger LOGGER = Logger.getLogger("com.ocient.jdbc");
 
+	  public enum Tls {
+		  OFF,    // No TLS
+		  UNVERIFIED, // Don't verify certificates
+		  ON,     // TLS but no server identity verification
+		  VERIFY, // TLS with server identity verification
+	  }
```
jgithub, line 85: TLS variants

daveboutcher: > TLS variants

Not really...just how we deal with it.

PR #55, opened 2020-08-27
Fix forced infinite loop.

tli-ocient:
https://jira.ocient.com:8443/browse/DB-12725

### DIFF
--- a/src/main/java/com/ocient/jdbc/XGConnection.java
+++ b/src/main/java/com/ocient/jdbc/XGConnection.java
@@ -447,6 +447,7 @@ public class XGConnection implements Connection {
   }
 
   private void connect(final String ip, int port) throws Exception {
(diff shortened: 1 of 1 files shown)


src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1387,7 +1391,7 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
 			}
 
 			try {
-				clientHandshake(user, pwd, database, true);
+				clientHandshake(user, pwd, database, shouldRequestVersion);
```
tli-ocient, line 1394: This is the culprit of the infinite loop when force is used.
clientHandshake(...,true);
fetchserverversion()
XGStatement(this,false);
conn.copy(false);
retval.reconnect(false);
clientHandshake(..., true); which should be using shouldRequestVersion = false.

PR #56, opened 2020-08-29
[DB-12725] fix not reconnect

tli-ocient:
https://jira.ocient.com:8443/browse/DB-12725

Fix for jdbc driver not reconnecting.

Now you can connect. Then do something like `pkill -9 rolehostd` . Then relaunch the server. The client will try to reconnect every time you type a command. You still have to wait for the server to come back up though. Be patient.

### DIFF
--- a/src/main/java/com/ocient/jdbc/XGStatement.java
+++ b/src/main/java/com/ocient/jdbc/XGStatement.java
@@ -1182,6 +1182,7 @@ public class XGStatement implements Statement {
 
 	private Object sendAndReceive(String sql, final Request.RequestType requestType, final int val,
 			final boolean isInMb, final Optional<Function<Object, Void>> additionalPropertySetter) throws SQLException {
+		LOGGER.log(Level.INFO, "Entered sendAndReceive()");
 		clearWarnings();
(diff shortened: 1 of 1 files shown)


src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -1337,8 +1339,9 @@ private Object sendAndReceive(String sql, final Request.RequestType requestType,
 				}
 
 				return br;
-			} catch (SQLException | IOException e) {
+			} catch (SQLException | NullPointerException |IOException e ) {
 				if (e instanceof SQLException && !SQLStates.UNEXPECTED_EOF.equals((SQLException) e)) {
```
tli-ocient, line 1343: Theres a line above `conn.out.write(intToBytes(wrapper.getSerializedSize()));` 1309. 
When we are disconnected, we will try to reconnect once. Then it will stop trying to reconnect every subsequent time. But that is only because something in the first reconnect attempt causes one of the things that line 1309 attempts to access to become a null pointer. Adding this exception here will make us attempt to reconnect every time.

PR #57, opened 2020-08-30
Keep PlanProtocol.proto in sync

daveboutcher:


jasonar81:
We don't have to worry about this anymore. We are about to remove plan proto from the JDBC driver.

PR #60, opened 2020-09-05
fix jdbc source file

tli-ocient:
Fix JDBC source.


src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1203,7 +1203,7 @@ private static boolean sourceCommands(BufferedReader reader) throws IOException
 						String cont = reader.readLine();
 						if (cont == null)
 							return quit;
-
+						line = cont;
```
tli-ocient, line 1206: If we don't do this, then we scrub on the first line until the reader hits the last line.

jasonar81 (approved):

PR #62, opened 2020-09-13
[DB-12845] output_next_query for explain and others

tli-ocient:
https://jira.ocient.com:8443/browse/DB-12845

Ugly diff, but really we are just adding an 
`if(outputfile){
print to output file;
}
else {
old code;
}
`

### DIFF
--- a/src/main/java/com/ocient/jdbc/XGStatement.java
+++ b/src/main/java/com/ocient/jdbc/XGStatement.java
@@ -1801,13 +1801,9 @@ public class XGStatement implements Statement {
 		}
 		String pm = explainPlan(plan, format);
 		ArrayList<Object> rs = new ArrayList<>();
-		String lines[] = pm.split("\\r?\\n");
-		for (int i = 0; i < lines.length; i++) {
-			String str = lines[i];
-			ArrayList<Object> row = new ArrayList<>();
-			row.add(str);
-			rs.add(row);
-		}
+		ArrayList<Object> row = new ArrayList<>();
+		row.add(pm);
+		rs.add(row);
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -1801,13 +1801,9 @@ private ResultSet explainPlanSQL(final String cmd) throws SQLException{
 		}
 		String pm = explainPlan(plan, format);
 		ArrayList<Object> rs = new ArrayList<>();
-		String lines[] = pm.split("\\r?\\n");
-		for (int i = 0; i < lines.length; i++) {
-			String str = lines[i];
-			ArrayList<Object> row = new ArrayList<>();
```
tli-ocient, line 1807: Is there a reason we were breaking up the string by new line characters and putting them into separate lines of the result set? It doesn't make a difference when its printed to stdout and is a bit uglier when printed to an output file.

jasonar81: I believe we had to do it that way for dbeaver to work.

jasonar81 (changes requested):
Can you bump the version in pom.xml and add a release note.

src/main/java/com/ocient/jdbc/XGStatement.java
tli-ocient: It will be fine like this. They will go into one result cell. That makes it easier to copy and paste. Also Dbeaver has this value thing on the right side where you can view it easily.
Screen Shot 2020-09-04 at 10 02 08 PM

jasonar81 (approved):

PR #67, opened 2020-09-21
[DB-12867] fix jdbc not reconnecting

tli-ocient:
https://jira.ocient.com:8443/browse/DB-12867

Simulated situation on ticket like so:
1. Run a long running query.
2. Kill server once result header is returned.
3. Bring the server back up.
4. Test if reconnects.

### DIFF
--- a/src/main/java/com/ocient/jdbc/XGStatement.java
+++ b/src/main/java/com/ocient/jdbc/XGStatement.java
@@ -855,6 +855,7 @@ public class XGStatement implements Statement {
 	private ClientWireProtocol.FetchSystemMetadataResponse.Builder fetchSystemMetadata(
 			final FetchSystemMetadata.SystemMetadataCall call, final String schema, final String table,
 			final String col, final boolean test) throws SQLException {
+		LOGGER.log(Level.INFO, "Entered fetchSystemMetadata()");
 		clearWarnings();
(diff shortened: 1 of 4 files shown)


src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -905,11 +906,15 @@ public String exportTranslation(final String table) throws SQLException {
...
 				return br;
-			} catch (SQLException | IOException e) {
+			} catch (SQLException | NullPointerException | IOException e) {
 				LOGGER.log(Level.WARNING, "fetchSystemMetadataResponse: ", e);
 				if (e instanceof SQLException && !SQLStates.UNEXPECTED_EOF.equals((SQLException) e)) {
 					throw e;
 				}
+				if (e instanceof IOException){
+					// When its a socket exception like this, the query is not cancelled.
+					setQueryCancelled(false);
```
tli-ocient, line 916: In our previous fix: https://github.com/Xeograph/ocient_jdbc/pull/56
we fixed the situation where if the server is brought down, we reconnect. However, that did not include the situation where the server is brought down in the middle of an open result set. In this situation, a query cancel is issued, which flips our flag for queryCancelled to true. In that case, we skip over the reconnect after it. We can unflip that flag here.

jasonar81 (approved):

PR #74, opened 2020-10-30
Add DDL statement for INVALIDATE STATS

ejiang-Ocient:


tli-ocient (approved):
nice! As long as it works by going through the vanilla update path.

PR #76, opened 2020-11-04
DB-13593 Make SQLException codes unique

awagenmaker-Ocient:


leogreca:
Looks good. Just one thing missing:
- U need increment the version in pom.xml

PR #79, opened 2020-11-10
DESCRIBE TABLE uses SQL type names for arrays

apark-ocient:
Closes DB-13576

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>com.ocient</groupId>
 	<artifactId>ocient-jdbc4</artifactId>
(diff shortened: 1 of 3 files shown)


apark-ocient:
example output before:
```
int8_array_col2 (ARRAY(BYTE))
int16_array_col (ARRAY(SHORT))
int32_array_col (ARRAY(INT))
int64_array_col (ARRAY(LONG))
```
now:
```
int8_array_col2 (ARRAY(BYTE))
int16_array_col (ARRAY(SMALLINT))
int32_array_col (ARRAY(INT))
int64_array_col (ARRAY(BIGINT))
```

leogreca:
Why not the fix that that in the SQL node? That info should come from Request_RequestType_FETCH_METADATA
If you fix in the SQL node, you will fix for all drivers.

PR #80, opened 2020-11-17
Fix reconnect logic

tli-ocient:
https://jira.ocient.com:8443/browse/DB-13805

Alright I think this is what we want. Right before we reconnect, check the previously connected port and ip. If we cannot connect to it, then it's likely we just killed it with a query. Then don't resend. Otherwise, resend.

### DIFF
--- a/src/main/java/com/ocient/jdbc/XGConnection.java
+++ b/src/main/java/com/ocient/jdbc/XGConnection.java
@@ -247,7 +247,9 @@ public class XGConnection implements Connection {
 	protected final String url;
 	protected String ip;
 	protected String originalIp;
+	protected String connectedIp;
 	protected int originalPort;
+	protected int connectedPort;
 	protected String user;
 	protected String database;
 	protected String client = "jdbc";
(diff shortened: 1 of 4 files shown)


src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -459,60 +463,82 @@ public void connect() throws Exception {
   private void connect(final String ip, int port) throws Exception {
 	LOGGER.log(Level.INFO,String.format("Trying to connect to IP: %s at port: %d", ip, port));
     try {
-      switch (this.tls) {
-      case OFF:
-		LOGGER.log(Level.INFO,"Unencrypted connection");
```
tli-ocient, line 464: This huge block is just fixing formatting

leogreca (approved):

src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -1451,8 +1451,15 @@ private Object sendAndReceive(String sql, final Request.RequestType requestType,
 					setQueryCancelled(false);
 				}
 				passUpCancel(false);
+				// Something went wrong. Did we kill the server with this query? If we did, then don't resend.
+				boolean safeToReSend = conn.isSockConnected();
```
leogreca, line 1455: if isSockConnected() returns false why should we reconnect in the line below?

tli-ocient: This function only tests the previously connected port and host. localhost:4051 for example. Reconnect tries all the sql nodes. localhost:4015, localhost:4052, etc... . If we killed the previously connected host, we should not resent.

leogreca (approved):

PR #81, opened 2020-11-30
DB-13546: Implemented completed queries virtual table

jwilliams-ocient:

PR #82, opened 2020-12-02
Don't print stack trace when select queries complete exceptionally

jwilliams-ocient:
This was something I added to aid with debugging and was not intended to be merged with #81 !

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>com.ocient</groupId>
 	<artifactId>ocient-jdbc4</artifactId>
-	<version>1.62</version> <!--JDBC VERSION NUMBER HERE -->
+	<version>1.63</version> <!--JDBC VERSION NUMBER HERE -->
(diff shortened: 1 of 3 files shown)


.project
```diff
@@ -20,4 +20,15 @@
...
+		<filter>
+			<id>1606839504837</id>
+			<name></name>
+			<type>30</type>
+			<matcher>
+				<id>org.eclipse.core.resources.regexFilterMatcher</id>
+				<arguments>node_modules|.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
+			</matcher>
+		</filter>
+	</filteredResources>
```
leogreca, line 33: What is that for?

jwilliams-ocient: Wow that’s what I get for using `git add -u`. I should add this to my .gitignore

jwilliams-ocient: Will fix after dinner

tli-ocient (approved):
Just getting rid of stack trace print

leogreca (approved):

PR #84, opened 2020-12-09
Make completed queries vTable consistent with the rolehostd logs

jwilliams-ocient:
DB-13546: Implement completed queries virtual operator


tli-ocient (approved):
Looks fine to me. Please make sure we are keeping xgsrc updated with the latest versions, that the clientwire protos are all synced up, and if you are not making these changes for ODBC then make tickets to track them.

jwilliams-ocient:
Great! We're merging this PR to `db` today which will give us the version bump - is that sufficient or do you think I should create another xgsrc PR?

tli-ocient:
Well, if I am not mistaken, your PR to db will fail until 30 minutes after this PR merges. Reason the jar takes about 30 minutes to deploy and the xgsrc PR will try to get a jar that is not there. I could be wrong.

PR #95, opened 2021-01-28
Extract query id from EXECUTE_PLAN and EXECUTE_INLINE_PLAN responses

jwilliams-ocient:
Storing the query id allows the client to cancel/kill queries


jasonar81:
I will take care of redoing this PR.

jasonar81:
Redone as part of connection/statement pooling PR.

PR #100, opened 2021-02-05
db-15101-fix fetch server versions

tli-ocient:


tli-ocient:
https://jira.ocient.com:8443/browse/DB-15101

PR #104, opened 2021-02-09
add more logging and fix schema on copied fethserverversion connection

tli-ocient:
Will put this together with Ryan's matrix state additions for 1.78

### DIFF
--- a/src/main/java/com/ocient/jdbc/XGStatement.java
+++ b/src/main/java/com/ocient/jdbc/XGStatement.java
@@ -192,11 +192,12 @@ public class XGStatement implements Statement
 				{
 				}
 			}
-
+			LOGGER.log(Level.INFO, "newXGStatement returning a cached statement");
 			return retval;
 		}
 		else
 		{
+			LOGGER.log(Level.INFO, "newXGStatement returning a new statement");
 			return new XGStatement(conn, shouldRequestVersion);
 		}
 	}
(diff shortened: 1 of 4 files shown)


src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -608,6 +612,16 @@ public void close() throws SQLException
...
+				if(conn.setSchema.equals("")){
+					LOGGER.log(Level.INFO, "This connection has an empty schema.");
+					conn.setSchema = conn.getSchema();
+					LOGGER.log(Level.INFO,String.format("After correcting incorrect schema. setSchema: %s." , conn.setSchema));
+				}
+				if(conn.defaultSchema.equals("")){
+					LOGGER.log(Level.INFO, "This connection has an empty default schema.");
+					conn.defaultSchema = conn.setSchema;
+					LOGGER.log(Level.INFO,String.format("After correcting incorrect default schema. defaultSchema: %s", conn.defaultSchema));
+				}
```
tli-ocient, line 624: Not my favorite fix, but this should only happen once. When the first connection is created, that connection will not have a server version and empty setSchema and default schema. It will copy that connection and fetch the server version. When that connection is returned to the pool, its reset method will set the server side connection to empty if we don't fix this here.

jasonar81 (approved):

PR #106, opened 2021-02-11
[DB-14606] Explain Pipeline

wantan244:
add support for new ddl statement, EXPLAIN PIPELINE


tli-ocient (changes requested):
some small changes:
I think you need to add the explain pipeline case to xGStatemant::execute().
I think we will release this as 1.79, so this can stay the same in the pom.xml.
Update the notes in releasenotes.adoc for this change and Dan's change please.
For Dan's change, just say "Add spotbugs"

tli-ocient (approved):

PR #108, opened 2021-02-12
DB-15243 tuples serialized as 1 column

ems236:
DB changes + this will break a ton of tests that I haven't fixed yet
Sort of a WIP because I haven't figured out how to make this work in CSuite, but I don't think I'll need to change anything

See DB changes https://github.com/Xeograph/xgsrc/pull/15566


src/main/java/com/ocient/jdbc/XGArray.java
```diff
@@ -135,53 +137,7 @@ public int getBaseType() throws SQLException
...
-			case 19:
-				return "DATE";
-			case 20:
-				return "TIMESTAMP";
-			case 21:
-				return "TIME";
-			default:
-				throw SQLStates.INVALID_COLUMN_TYPE.clone();
-		}
+		return XGResultSetMetaData.type2Name(type);
```
leogreca, line 140: Thanks for that!

leogreca (approved):

PR #109, opened 2021-02-12
Remove extra NODE_NOT_FOUND

markwschmidt:


tli-ocient:
Thanks a lot for fixing this. While we are here, could you please also update release notes adoc for 1.80 / 1.81 as well? We need to keep that up to date. Small note is fine for example
1.80: Add cluster not found sql state
1.81: Fix cluster not found sql state compile error.

Thanks

markwschmidt:
You got it

PR #111, opened 2021-02-16
fix a bunch of OWASP bugs

tli-ocient:
This fixes a big chunk of the OWASP bugs. There are some that I thought were false positives or merits some discussion before fixing. I attach the new html bug report and my observations on each of the bugs.

See the ticket for those two new files: https://jira.ocient.com:8443/browse/DB-15309


src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1737,6 +1787,7 @@ private static void select(final String cmd)
 	{
 		long start = 0;
 		long end = 0;
+		System.out.println(String.format("Timmy debug: in select %s", cmd));
```
ems236, line 1790: This probably isn't meant to be here

tli-ocient: true , good catch

PR #130, opened 2021-04-06
[DB-16200] Improve explains by making JSON default and adding debug option

jlchen0:
Corresponding xgsrc PR:
https://github.com/Xeograph/xgsrc/pull/16486

### DIFF
--- a/src/main/java/com/ocient/jdbc/XGStatement.java
+++ b/src/main/java/com/ocient/jdbc/XGStatement.java
@@ -743,7 +743,7 @@ public class XGStatement implements Statement
 		try
 		{
 			passUpCancel(true);
-			if (sql.toUpperCase().startsWith("SELECT ") || sql.toUpperCase().startsWith("WITH ") || sql.toUpperCase().startsWith("EXPLAIN ") || sql.toUpperCase().startsWith("LIST TABLES")
+			if (sql.toUpperCase().startsWith("SELECT ") || sql.toUpperCase().startsWith("WITH ") || sql.toUpperCase().startsWith("EXPLAIN") || sql.toUpperCase().startsWith("LIST TABLES")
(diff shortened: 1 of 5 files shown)


src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -1188,15 +1188,28 @@ private ResultSet explainSQL(final String cmd) throws SQLException
 		ClientWireProtocol.ExplainFormat format;
 		String sql = "";
 
-		if (startsWithIgnoreCase(cmd, "EXPLAIN JSON "))
+		if (startsWithIgnoreCase(cmd, "EXPLAIN DEBUG"))
```
ems236, line 1191: I think the space actually matters a bit here because you're trimming the length including the space a few lines later.

ems236: It might be good to rewrite these in a way where whitespace doesn't matter anyway. I think Timmy is already looking at something like that for describe table, so check with him.

ems236 (approved):
Can you also add these changes to release_notes.adoc?
I think db is also on 1.98 now, so bump that to 1.99.

src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -1188,15 +1188,28 @@ private ResultSet explainSQL(final String cmd) throws SQLException
...
+			format = ClientWireProtocol.ExplainFormat.DEBUG; 
+			sql = cmd.substring("EXPLAIN DEBUG".length()).trim(); 
 		}
-		else
+		else if (startsWithIgnoreCase(cmd, "EXPLAIN PROTO"))
 		{
 			format = ClientWireProtocol.ExplainFormat.PROTO;
-			sql = cmd.substring("EXPLAIN ".length()).trim();
+			sql = cmd.substring("EXPLAIN PROTO".length()).trim(); 
+		}
```
tli-ocient, line 1200: same here.

tli-ocient (changes requested):
please check that the spacing is working correctly

tli-ocient (approved):

PR #132, opened 2021-04-14
[DB-16081] result set support + cache limit warning

tli-ocient:
https://ocient.atlassian.net/browse/DB-16081

### DIFF
--- a/src/main/java/com/ocient/jdbc/XGResultSet.java
+++ b/src/main/java/com/ocient/jdbc/XGResultSet.java
@@ -68,7 +68,7 @@ public final class XGResultSet implements ResultSet
 				attachToQuery(newConn, queryId);
 
 				// First fetch has to come from main thread
-				while (!didFirstFetch.get())
+				while (!didFirstFetch.get() && !cacheLimitBreak.get())
 				{
 					Thread.sleep(1);
 				}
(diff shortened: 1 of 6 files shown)


src/main/java/com/ocient/jdbc/XGResultSet.java
```diff
@@ -85,22 +85,83 @@ public void run()
...
+				LOGGER.log(Level.INFO, "First cache timeout limit hit. Asynchrnously finishing fetch on main resultSetThread.");
+				try
+				{
+					XGStatement cachedOrNewStatement = XGStatement.newXGStatement(conn, false);
+
+					// Take the old connection. It will be used to finish the fetch.
+					XGConnection oldConn = conn;
+					// Give the userstatement the new connection.
+					stmt.conn = cachedOrNewStatement.conn;
+					// The cachedOrNewStatement will just be thrown out.
```
jasonar81, line 142: Isn't there a way to do this such that when we are done with this, it will go back to the pool?

src/main/java/com/ocient/jdbc/XGResultSet.java
```diff
@@ -2814,7 +2913,9 @@ private boolean mergeData(final ClientWireProtocol.ResultSet re) throws Exceptio
 			didFirstFetch.set(true);
 		}
 
-		if (done && didFirstFetch.get())
+		// Need to not join threads here if we got the special break. The main thread cannot
+		// cannot return yet. It needs to finish the fetch.
+		if (done && didFirstFetch.get() && !cacheLimitBreak.get())
```
jasonar81, line 2918: Not sure I follow this. One of the key points is that we are not tying up the main user thread and that it can go submit a new query. Does this prevent that?

jasonar81 (changes requested):

src/main/java/com/ocient/jdbc/XGResultSet.java
tli-ocient: added returning that statement back to the pool

src/main/java/com/ocient/jdbc/XGResultSet.java
tli-ocient: No, a new query will come with a new result set.And I've tested this on the CLI. After issuing a query that hits the first limit, I can still fire off new queries as the other query finishes asynchronously

jasonar81 (approved):

jasonar81 (approved):

PR #146, opened 2021-05-19
Moved KML from CLI to driver.  Added different colors for GIS types and description with non-gis types in the same row.

raychen323:


src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -98,6 +100,21 @@ public void run()
...
+		{
+			KML.setKML(cmd.substring("OUTPUT GIS KML ".length()).trim());
+			if (KML.KMLIsEmpty())
+			{
+				System.out.println("Provide a filename to output the query to");
+			}
+		}
+		catch (final Exception e)
+		{
+			System.out.println("CLI Error: " + e.getMessage());
```
tli-ocient, line 115: Driver cannot have printouts. Change these to trace logs.

src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -966,6 +983,11 @@ else if (numClientThreads > Runtime.getRuntime().availableProcessors())
 
 			LOGGER.log(Level.INFO, "Creating result set for query with " + numClientThreads + " result set threads");
 			result = conn.rs = new XGResultSet(conn, fetchSize, this, numClientThreads);
+
+			if(!KML.KMLIsEmpty()) {
+				KML.outputGeospatial(result, result.getMetaData());
```
tli-ocient, line 988: Since you can get the metadata from the result set anyway, theres no reason to have this function take the result set and the result set metadata. Just have it call getMetaData inside outputGeoSpatial. I know this is what the CLI was doing, but thats old code.

tli-ocient (changes requested):
Nice work. Just some small comments. Also, if you haven't already, can you do some local testing?

Run a few of these KML's, then make sure queries are still working after you output a few files.

PR #157, opened 2021-06-23
[DB-17148] fix timoutMillis again

tli-ocient:
https://ocient.atlassian.net/browse/DB-17148

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.ocient</groupId>
   <artifactId>ocient-jdbc4</artifactId>
-  <version>2.14</version>
+  <version>2.15</version>
   <!--JDBC VERSION NUMBER HERE -->
   <name>${project.groupId}:${project.artifactId}</name>
   <description>JDBC Driver for connecting to an Ocient Database</description>
--- a/userdoc/release_notes.adoc
+++ b/userdoc/release_notes.adoc
@@ -5,6 +5,13 @@ All our jdbc drivers are located {drivers_repo}[here].
 Below is the release notes for every driver version that has customer implication.
 
 //tag::compact[] 
+== 2.15 (2021-6-21)
+
+New Features:: 
+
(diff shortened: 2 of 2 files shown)


src/main/java/com/ocient/jdbc/XGResultSet.java
```diff
@@ -2856,6 +2869,22 @@ else if (type == 24) // ST_POLYGON
...
 	}
 
+	/*
+	 * Checks if the result set has actual data or if it is a 0 row ping buffer.
+	 * Returns true for actual data, false for the ping buffer. We will also
+	 * consider a dem buffer actual data.
+	 */
+	private boolean receivedActualData(final ClientWireProtocol.ResultSet resultSet){
+		final List<ByteString> buffers = resultSet.getBlobsList();
+		// If the first buffer has 0 rows, then it is a zero ping buffer. The number of buffers should always be greater than 0.
```
jasonar81, line 2879: This isn't true if its an error response. You aren't checking that first.

src/main/java/com/ocient/jdbc/XGResultSet.java
```diff
@@ -1508,23 +1508,36 @@ private void getMoreData(final XGConnection newConn)
 					LOGGER.log(Level.WARNING, "Thread saw cacheLimitBreak. Exiting getMoreData");
 					return;
 				}
-				newConn.out.write(intToBytes(wrapper.getSerializedSize()));
-				wrapper.writeTo(newConn.out);
-				newConn.out.flush();
 
 				// Kind of ugly, but doesn't violate JMM (startTask() is synchronous)
 				final ClientWireProtocol.FetchDataResponse.Builder fdr = ClientWireProtocol.FetchDataResponse.newBuilder();
+				final long rpcTimeout = getTimeoutMillis();
```
jasonar81, line 1514: I think it makes more sense to just put the whole while(true) loop in the task and (other than that) leave the code exactly like it was before.

jasonar81 (changes requested):

PR #161, opened 2021-06-26
DB-16761

jasonar81:


src/main/java/com/ocient/cli/CLI.java
```diff
@@ -2138,43 +2161,54 @@ private static boolean sourceCommands(final BufferedReader reader) throws IOExce
...
 					if (!comment && quote == '\0' && cmd.trim().endsWith(";"))
 					{
+						// System.out.println("Finished scrubbing command: '" + cmd + "'");
 						cmd = cmd.trim();
 						cmd = cmd.substring(0, cmd.length() - 1).trim();
 						if (trace && !endsWithIgnoreCase(cmd, " trace"))
 						{
 							cmd = cmd + " trace";
 						}
+						// System.out.println("Finished trimming scrubbed: '" + cmd + "'");
```
tli-ocient, line 2177: Did you mean to clean this up?

tli-ocient (changes requested):
I don't think you need to do another version bump since we didnt put 2.18 into db yet.

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -2149,32 +2173,41 @@ private static boolean sourceCommands(final BufferedReader reader) throws IOExce
...
 							cmd = cmd + " trace";
 						}
+
 						break;
 					}
 					else
 					{
-						final String cont = reader.readLine();
-						if (cont == null)
+						// System.out.println("Current command text: '" + cmd + "'");
```
tli-ocient, line 2181: this too

src/main/java/com/ocient/cli/CLI.java
jasonar81: Done

src/main/java/com/ocient/cli/CLI.java
jasonar81: Done

PR #168, opened 2021-07-11
[DB-17360] have set schema handle caps

tli-ocient:
https://ocient.atlassian.net/browse/DB-17360


src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -2999,6 +2999,8 @@ private int setSchema(final String cmd) throws SQLException
 			}
 
 			schema = schema.substring(1, schema.length() - 1);
+		} else {
+			schema = schema.toLowerCase();
 		}
```
tli-ocient, line 3004: If unquoted, lowercase. If quoted, leave alone

jasonar81 (approved):

PR #176, opened 2021-08-30
[DB-17875] sso handshake support 

tli-ocient:
https://ocient.atlassian.net/browse/DB-17875

### DIFF
--- a/src/main/java/com/ocient/jdbc/XGConnection.java
+++ b/src/main/java/com/ocient/jdbc/XGConnection.java
@@ -2893,7 +2893,8 @@ public class XGConnection implements Connection
 		innerBuilder.setConcurrency(parallelism != null ? parallelism : 0);
 		builder.setConcurrency(innerBuilder.build());
 
-		int rowsModified = sendParameterMessage(builder.build());		
+		int rowsModified = sendParameterMessage(builder.build());
+		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.
 		if(reset){
 			this.parallelism = null;
 		} else {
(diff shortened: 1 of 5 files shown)


jwilliams-ocient:
Looks good! My only question is what is the behavior when using this handshake on a remote machine (e.g. dev container)?

tli-ocient:
The behavior is the handshake will attempt to happen, but will (likely) fail because `isDeskTopSupported()` will fail. This is due to the environment not supporting `java.awt.Desktop` . If it does though, then it would work. If it doesn't, it will just print an error saying so.

jwilliams-ocient:
I think that behavior is fine for the Driver itself, but for the CLI we should print the URL to stdout so the user can click on it. Is that possible?

jwilliams-ocient:
I'd like this to be the primary way that devs connect to remote clusters like battleship

tli-ocient:
yes, easy.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -969,9 +1104,111 @@ else if (oneShotForce)
...
+			Desktop desktop = Desktop.getDesktop();
+			try{
+				desktop.browse(new URI(authUrl));
+			} catch (Exception e) {
+				LOGGER.log(Level.WARNING, String.format("Failed to open browser for URI: %s with Desktop", authUrl));
+				throw e;
+            }
+		} else {
+			LOGGER.log(Level.WARNING, String.format("Could not open default browser with Desktop library. Please proceed to the following url on a browser: %s", authUrl));
+			System.out.println(String.format("Could not open default browser with Desktop library. Please proceed to the following url on a browser: %s", authUrl));
```
tli-ocient, line 1129: I am a bit hesitant to put print statements in the actual driver code, but it seems like the best way to do what Jordan is describing. We cannot throw an exception here because it would end the handshake . So we will both log and print the url and whoever is using the driver should be able to proceed to the url. Is this OK? 
@jasonar81

jasonar81 (approved):

PR #181, opened 2021-09-21
DB-18257: Process token signature and issuer fingerprint in SSO handshake

jwilliams-ocient:


tli-ocient (changes requested):
please see my comments in slack

jwilliams-ocient:
Will use #180

PR #184, opened 2021-09-29
generate and upload userdocs as part of CI

tli-ocient:
https://ocient.atlassian.net/browse/DB-18204

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,9 @@
     </snapshotRepository>
   </distributionManagement>
   <properties>
+    <asciidoctor-maven-plugin.version>2.1.0</asciidoctor-maven-plugin.version>
+    <asciidoctorj.pdf.version>1.6.0</asciidoctorj.pdf.version>
+    <asciidoctorj.version>2.5.2</asciidoctorj.version>  
     <protoc>${project.basedir}/bazel-protoc.sh</protoc>
     <spotbugs-maven-plugin.version>4.2.0</spotbugs-maven-plugin.version>
     <findsecbugs-plugin.version>1.10.1</findsecbugs-plugin.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -39,6 +39,9 @@
     </snapshotRepository>
   </distributionManagement>
   <properties>
+    <asciidoctor-maven-plugin.version>2.1.0</asciidoctor-maven-plugin.version>
+    <asciidoctorj.pdf.version>1.6.0</asciidoctorj.pdf.version>
+    <asciidoctorj.version>2.5.2</asciidoctorj.version>
```
rveitch-Ocient, line 44: This property isn't used anywhere?

pom.xml
```diff
@@ -95,6 +98,60 @@
...
+        <dependencies>
+            <dependency>
+                <groupId>org.asciidoctor</groupId>
+                <artifactId>asciidoctorj-pdf</artifactId>
+                <version>${asciidoctorj.pdf.version}</version>
+            </dependency>
+        </dependencies>        
+        <executions>
+          <execution> 
+              <id>output-html</id>
```
rveitch-Ocient, line 114: Small nit, would be good to give each execution consistently named ids, right now they're `output-html` and `generate-pdf-doc`. The names themselves are arbitrary. How about `build-html-release-notes` and `build-pdf-release-notes`? Otherwise something that is consistent across both.

pom.xml
```diff
@@ -95,6 +98,60 @@
...
+                  <goal>process-asciidoc</goal>
+              </goals>
+              <configuration>
+                  <backend>pdf</backend>
+                  <attributes>
+                      <source-highlighter>coderay</source-highlighter>
+                      <icons>font</icons>
+                      <pagenums/>
+                      <idprefix/>
+                      <idseparator>-</idseparator>
```
rveitch-Ocient, line 139: Did you validate that these attributes are what you want? A lot of the time this kind of config is embedded in the source asciidoc front matter but in this case I don't see any front matter in release_notes.adoc

pom.xml
```diff
@@ -95,6 +98,60 @@
...
+                  </attributes>
+              </configuration>
+          </execution>
+        </executions>        
+        <configuration>
+          <sourceDirectory>userdoc</sourceDirectory>
+          <outputDirectory>public</outputDirectory>
```
rveitch-Ocient, line 146: Did you want to explicitly config `public` as your output directory? If not you could always use the default output directory of `target/generated-docs`

rveitch-Ocient (changes requested):

pom.xml
tli-ocient: right, removed

pom.xml
tli-ocient: good idea, changed

pom.xml
tli-ocient: no, I did not need those

pom.xml
tli-ocient: Ok, the default works fine

rveitch-Ocient (approved):
LGTM

PR #192, opened 2021-10-29
add handling for refreshing with an expired token

tli-ocient:


src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1176,9 +1189,16 @@ private void clientHandshakeSecurityToken(final String db, final boolean shouldR
 			throw e;
 		}
 		catch (final Exception e)
-		{
-			LOGGER.log(Level.WARNING, String.format("Exception %s occurred during handshake with message %s", e.toString(), e.getMessage()));
-
+		{	
+			// TODO Timmy: Change this to the right corrected state.
+			if(e instanceof SQLException && SQLStates.SESSION_EXPIRED.equals((SQLException) e)){
+				LOGGER.log(Level.INFO, "clientHandshakeSecurityToken() received session expired. Attempting to refresh session");
```
tli-ocient, line 1195: @jwilliams-ocient just let me know what this expected state needs to be. Right now I have it set as `SESSION_EXPIRED` which is not technically correct.

PR #198, opened 2021-11-08
[SOL-12] DataExtract-ExtractConfiguration

tli-ocient:
https://ocient.atlassian.net/browse/SOL-12

Let me know if there are some other unit tests I should add.


src/test/java/com/ocient/cli/extract/ExtractConfigurationTest.java
```diff
@@ -0,0 +1,103 @@
...
+import junitparams.Parameters;
+import static org.junit.Assert.assertEquals;
+
+import java.util.Properties;
+
+@RunWith(JUnitParamsRunner.class)
+public class ExtractConfigurationTest {
+
+    @Test
+    public void test_defaultConfig()
```
cj81499, line 19: nit, remove `test_` prefix from all tests

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,202 @@
...
+            .collect(
+                Collectors.toUnmodifiableMap(k -> k.toLowerCase(), config::getProperty)
+            )
+        );    
+    }
+
+    // For certain configuration options, additional checks are necessary.
+    private void validateConfiguration() throws ParseException{
+
+        if(!locationType.equals("LOCAL") && !locationType.equals("S3")){
```
cj81499, line 182: Each of `LOCAL`/`S3`, `DELIMITED`, `NONE`/`GZIP`, and `S3`/`NONE` should be an `enum`. This is especially important b/c it's pretty likely that more options need to be added to each of these in the future (ex, additional compression types).

`EnumUtils` ([docs](https://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/EnumUtils.html)) may be helpful.

src/test/java/com/ocient/cli/extract/ExtractConfigurationTest.java
tli-ocient: ok, removed

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: good point, changed

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+    public static final String DEFAULT_FILE_EXTENSION = ".csv";
+
+    public static final String MAX_ROWS_PER_FILE = "max_rows_per_file";
+    public static final int DEFAULT_MAX_ROWS_PER_FILE = 0;
+
+    public static final String COMPRESSION = "compression";
+    public static final String DEFAULT_COMPRESSION = "NONE";
+
+    public static final String BUCKET = "bucket";
+    public static final String DEFAULT_BUCKET = "NONE";
```
dcoombs-ocient, line 76: Default should be `""`

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+    public enum Compression{
+        NONE,
+        GZIP;
+    }
+
+    // Defaults for all options and expected paths in the property map.
+    public static final String LOCATION_TYPE = "location_type";
+    
+    public static final String FILE_TYPE = "file_type";
+    public static final String DEFAULT_FILE_TYPE = "DELIMITED";
```
cj81499, line 61: Does it make more sense for this to be a `FileType` instead of a `String`?

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+
+    ExtractConfiguration(final Properties properties){
+        notNull(properties, "ExtractConfiguration received null properties object");
+        final ImmutableConfiguration config = createConfig(properties);
+
+        // Only location type is required.
+        locationType = LocationType.valueOf(config.getString(LOCATION_TYPE).toUpperCase());
+        notNull(locationType, "Location type is a required configuration option");
+        // To achieve case insensitivity for file type
+        fileType = FileType.valueOf(config.getString(FILE_TYPE, DEFAULT_FILE_TYPE).toUpperCase());
```
cj81499, line 25: Thoughts on using [EnumUtils.getEnumIgnoreCase](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/EnumUtils.html#getEnumIgnoreCase-java.lang.Class-java.lang.String-E-)?

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+    public static final String COMPRESSION = "compression";
+    public static final String DEFAULT_COMPRESSION = "NONE";
+
+    public static final String BUCKET = "bucket";
+    public static final String DEFAULT_BUCKET = "NONE";
+
+    public static final String AWS_KEY_ID = "aws_key_id";
+    public static final String DEFAULT_AWS_KEY_ID = "";
+
+    public static final String AWS_KEY_SECRET = "aws_key_secret";
```
dcoombs-ocient, line 81: This should be `aws_secret_key`

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+    public static final String DEFAULT_FILE_PREFIX = "results-";
+
+    public static final String FILE_EXTENSION = "file_extension";
+    public static final String DEFAULT_FILE_EXTENSION = ".csv";
+
+    public static final String MAX_ROWS_PER_FILE = "max_rows_per_file";
+    public static final int DEFAULT_MAX_ROWS_PER_FILE = 0;
+
+    public static final String COMPRESSION = "compression";
+    public static final String DEFAULT_COMPRESSION = "NONE";
```
cj81499, line 73: `Compression` instead of a `String`?

cj81499, line 73: `Compression` instead of a `String`?

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+import org.apache.commons.collections4.IteratorUtils;
+import org.apache.commons.configuration2.Configuration;
+import org.apache.commons.configuration2.ConfigurationConverter;
+import org.apache.commons.configuration2.ImmutableConfiguration;
+import org.apache.commons.configuration2.MapConfiguration;
+import static org.apache.commons.lang3.Validate.notNull;
+
+import com.ocient.cli.ParseException;
+
+public class ExtractConfiguration {
```
dcoombs-ocient, line 15: Can we keep the formatting of this file inline with the formatting of the rest of the driver (braces all on newlines)?

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+            IteratorUtils.toList(config.getKeys()).stream()
+            .collect(
+                Collectors.toUnmodifiableMap(k -> k.toLowerCase(), config::getProperty)
+            )
+        );    
+    }
+
+    // For certain configuration options, additional checks are necessary.
+    // Future validation logic can be added here as necessary.
+    private void validateConfiguration() throws ParseException{
```
dcoombs-ocient, line 195: Does this class have access to a logger at all? Could we set that up?

I'd like if this function also logged warnings when we see configuration that will be ignored. For example if you set a `BUCKET` but you are doing a `local` load.

dcoombs-ocient (changes requested):
Mostly LGTM, a few comments.

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,203 @@
...
+    private static ImmutableConfiguration createConfig(final Properties properties) {
+      return remapConfigKeys(ConfigurationConverter.getConfiguration(properties));
+    }
+    
+    // Helper function to implement case insensitivity by making everything lower case.
+    private static Configuration remapConfigKeys(final ImmutableConfiguration config){
+        return new MapConfiguration(
+            IteratorUtils.toList(config.getKeys()).stream()
+            .collect(
+                Collectors.toUnmodifiableMap(k -> k.toLowerCase(), config::getProperty)
```
cj81499, line 188: Can we use `String::toLowerCase` instead of `k -> k.toLowerCase()`?

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
cj81499, line 76: Why not just use `null`?

cj81499: Over `null`?

dcoombs-ocient: `null` is better

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: No, this is the value / default value pulled out of properties, which is then converted to an enum.

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: same as above

tli-ocient: same as above

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: yes

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: No, the difference between that and `valueOf` is that instead of throwing, it resolves to a null. I think we want to just throw here so that this fails immediately.

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: changed

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: changed to null

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: I changed it for this file... but the formatting is not consistent throughout the driver anyway. I would rather we not enforce this rule until we decide on a formatter for the driver as a whole. At that point we can apply it to all the files. It's hard for me to manually do this as I am used to typing it out this way (opening braces on the same line).

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: let me see

tli-ocient: added a logger and a warning

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
cj81499: As is, we're still using a String to represent an Enum for `DEFAULT_FILE_TYPE` (and `DEFAULT_COMPRESSION`).

If we change the type of `DEFAULT_FILE_TYPE` to `FileType`, we can use `getEnumIgnoreCase` with a `defaultEnum` argument. The code would look like:

```java
fileType = EnumUtils.getEnumIgnoreCase(FileType.class, config.getString(FILE_TYPE), DEFAULT_FILE_TYPE);
```

If you want to be consistent and use `getEnumIgnoreCase` for all of these, `notNull` returns the argument you pass it, so we can wrap the one used to get `locationType` with `notNull` like so:

```java
locationType  = notNull(EnumUtils.getEnumIgnoreCase(LocationType.class, config.getString(LOCATION_TYPE)), "Location type is a required configuration option");
```

but I'm also perfectly content sticking w/ `Enum.valueOf` for it, because there is no default.

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,235 @@
...
+{
+
+    public ExtractConfiguration(final Properties properties)
+    {
+        notNull(properties, "ExtractConfiguration received null properties object");
+        final ImmutableConfiguration config = createConfig(properties);
+
+        // Only location type is required.
+        locationType = LocationType.valueOf(config.getString(LOCATION_TYPE).toUpperCase());
+        notNull(locationType, "Location type is a required configuration option");
```
cj81499, line 28: Reading the  [docs](https://commons.apache.org/proper/commons-configuration/apidocs/index.html), it sounds like `config.getString` can return `null`, so `toUpperCase` could cause a NPE.

On the other hand, `Enum.valueOf` will throw `IllegalArgumentException` "if the specified enum type has no constant with the specified name" ([docs](https://docs.oracle.com/javase/7/docs/api/java/lang/Enum.html#valueOf(java.lang.Class,%20java.lang.String))).

I think `notNull` just needs to be moved to check the result of `config.getString(LOCATION_TYPE)`.

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: Maybe I am misunderstanding you.
I think doing this for `fileType`:
`fileType = EnumUtils.getEnumIgnoreCase(FileType.class, config.getString(FILE_TYPE), DEFAULT_FILE_TYPE);
`
would result in getting the default (DELIMITED), if they ever specify something incorrect (`file_type="badThing"`), rather than erroring loudly which is what I want.

If I do it without the default argument,
`fileType = EnumUtils.getEnumIgnoreCase(FileType.class, config.getString(FILE_TYPE));`
then that would result in null if `file_type` is unspecified because `config.getString(FILE_TYPE)` would return null

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: Good point. changed.

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
dcoombs-ocient: I didn't realize that the code was mixed-mode on this already. That's fine, we will add a formatter in the future, but for now you can chose whatever format you want for new files. For existing files try and keep the same mode if possible.

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
dcoombs-ocient: I definitely think we want to catch something like `file_type="bad_thing"`. `EnumUtils.getEnumIgnoreCase` with a default hides that, so we shouldn't use it.

Lets stick with the code as is. However, we should catch the exceptions thrown by `Enum.valueOf` (looks like `IllegalArgumentException` and `NullPointerException`) and wrap them in a `ParseException` such that the user of this constructor only needs to catch `ParseException`.

tli-ocient: done

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
cj81499: Rather than hardcoding the String, can we get it from `Compression.NONE.toString()`?

It's unlikely, but that way if `NONE` is no longer a valid compression type in the future, the compiler will yell at us. The same applies to `DEFAULT_FILE_TYPE`

dcoombs-ocient (approved):

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,242 @@
...
+        final ImmutableConfiguration config = createConfig(properties);
+
+        try{
+            // Only location type is required.
+            locationType = LocationType.valueOf(notNull(config.getString(LOCATION_TYPE), "Location type is a required configuration option").toUpperCase());
+            // To achieve case insensitivity for file type
+            fileType = FileType.valueOf(config.getString(FILE_TYPE, DEFAULT_FILE_TYPE).toUpperCase());
+            // To achieve case insensitivity for compression
+            compression = Compression.valueOf(config.getString(COMPRESSION, DEFAULT_COMPRESSION).toUpperCase());            
+            maxRowsPerFile = config.getInt(MAX_ROWS_PER_FILE, DEFAULT_MAX_ROWS_PER_FILE);
```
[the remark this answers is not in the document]
tli-ocient: possible conversion error. `geInt("Abcd")`

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -0,0 +1,242 @@
...
+    public static final String DEFAULT_FILE_PREFIX = "results-";
+
+    public static final String FILE_EXTENSION = "file_extension";
+    public static final String DEFAULT_FILE_EXTENSION = ".csv";
+
+    public static final String MAX_ROWS_PER_FILE = "max_rows_per_file";
+    public static final int DEFAULT_MAX_ROWS_PER_FILE = 0;
+
+    public static final String COMPRESSION = "compression";
+    public static final String DEFAULT_COMPRESSION = "NONE";
```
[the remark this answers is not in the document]
tli-ocient: yes, changed

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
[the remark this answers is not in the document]
cj81499: Makes sense. It looks like `getBoolean` (for `skipHeader`) can too ([docs](https://commons.apache.org/proper/commons-configuration/javadocs/v1.10/apidocs/org/apache/commons/configuration/Configuration.html#getBoolean(java.lang.String))), so we ought to move it.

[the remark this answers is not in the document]
tli-ocient: done, thanks

cj81499 (approved):
LGTM!

PR #199, opened 2021-11-09
[SOL-11] Add extract to syntax parser

dcoombs-ocient:


src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private static Properties parseOptions(String ops) throws ParseException
+    {
+        if (ops == null || ops.equals("")) 
+        {
+            return new Properties();
+        }
+        
+        ops = ops.replaceAll("\\s+","");
+        ops = ops.replaceAll("\\n","");
```
cj81499, line 93: First of all, I believe `\s+` should capture newlines.

Second, I don't think we can safely remove all whitespace, because `FILE_NAME`, `PATH`, `RECORD_DELIMITER`, and `FIELD_DELIMITER` (and maybe others) could contain meaningful whitespace. (The default `RECORD_DELIMITER` is a newline after all).

I think something like `\s*(?<key>.*)\s*=\s*(?<value>.*)\s*` on each pair might work?

There's also the problem of removing quotes from values. Perhaps `\s*(?<key>.*)\s*=\s*(?<quote>['"])(?<value>.*)\k<quote>\s*` could do the trick?

There's probably still more complexity to this than what I've come up with here.

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
```diff
@@ -0,0 +1,109 @@
...
+        Properties ret = new Properties();
+        for (String pair: pairs) 
+        {
+            String[] kv = pair.split("=");
+            if (kv.length != 2) 
+            {
+                throw new ParseException("Bad key-value, must be separated by '='" + pair);
+            }
+            ret.setProperty(kv[0], kv[1]);
+        }
```
cj81499, line 105: This silently overwrites a previous value when there is a repeat key.

We could fix this and use a more functional style by converting pairs into a `Stream` and mapping/collecting into a `Map<String, String>` (which will throw an `IllegalStateException` if there is a duplicate key ([docs](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-))), and then use `Properties.putAll`.

src/test/java/com/ocient/cli/extract/ExtractSyntaxParserTest.java
```diff
@@ -0,0 +1,191 @@
...
+    {
+        if (ops != null)
+        {
+            String opsStr = "OPTIONS(";
+            for (final Map.Entry<String,String> op: ops.entrySet())
+            {
+                opsStr += op.getKey() + delim + op.getValue() + eol;
+            }
+            opsStr = opsStr.substring(0, opsStr.length() - eol.length());
+            opsStr += ")";
```
cj81499, line 38: I know this is just test code, but...

`String` concatenation in a loop is a performance concern. Use a `StringBuilder`.

It also might be cleaner to use `ops.forEach((k, v) -> {...})`.

With that all said, I believe an even better approach is to convert `ops.entrySet()` to a `Stream` and use [`Collectors.joining(delimiter, prefix, suffix)`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#joining-java.lang.CharSequence-java.lang.CharSequence-java.lang.CharSequence-). That way, we avoid the `substring` "hack" to remove the trailing ",".

cj81499 (changes requested):
It's likely worth adding some test cases for the edge cases (whitespace handling and repeat keys) that I mentioned in the individual comments.

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1400,6 +1405,25 @@ private static void select(final String cmd)
...
+		catch (ParseException e)
+		{
+			System.out.println(e.getMessage());
+			return;
+		}
+
+		// TODO run the query and extract using the options
+		System.out.println("Query: " + res.getQuery());
+		System.out.println("Options: " + res.getConfig());
+		System.out.println("Not implemented");
```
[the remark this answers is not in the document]
dcoombs-ocient: That would crash the CLI. I'd have to catch this in the CLI main loop and print the same thing. Going to leave this as its getting replaced in a follow up PR.

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
```diff
@@ -0,0 +1,109 @@
...
+ * A utility class for parsing the "EXTRACT TO" syntax into its
+ * configuration and query components
+ */ 
+public class ExtractSyntaxParser
+{
+    private static String TYPE_GROUP = "type";
+    private static String OPTIONS_GROUP = "ops";
+    private static String QUERY_GROUP = "query";
+
+    private static final Pattern extractPattern = Pattern.compile(
```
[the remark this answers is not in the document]
dcoombs-ocient: Done

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
dcoombs-ocient: You're right that the `\n` replace was redundant. Removed

dcoombs-ocient: When I was playing around with this originally, this works:

```
Ocient> extract to local OPTIONS(delim="\n") as select * from table;
Query: select * from table
Options: {location_type=local, delim="\n"}
Not implemented
```

I definitely missed removing the quotes though.

I think the CLI is doing some escaping of special characters when it reads the command ing.

dcoombs-ocient: I hadn't considered spaces in paths/filenames, I'll need to fix that as well.

tli-ocient: agreed, we probably need a more robust solution than just removing all whitespaces.

dcoombs-ocient: Agreed, I missed a bunch here. Thanks for checking my logic @cj81499. Working on a more robust approach.

PR #200, opened 2021-11-12
User/tli/record extractor

tli-ocient:
https://ocient.atlassian.net/browse/SOL-13
https://ocient.atlassian.net/browse/SOL-21


tli-ocient:
excuse the merge conflict gore

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+public class ResultSetExtractor{
+
+    ResultSetExtractor(final ExtractConfiguration config){
+        extractConfig = config;
+        recordExtractorFactory = new RecordExtractorFactory(config);
+    }
+
+    // Extract the result set given its metadata and desired ExtractConfiguration.
+    public void extract(final ResultSet resultSet, final ResultSetMetaData resultSetMetaData) throws SQLException{
```
tli-ocient, line 26: @dcoombs-ocient I don't remember if you said you wanted __everything__ to throw ParseExceptions only. There are a bunch of operations here that could throw SQL exceptions. I think that might be more appropriate.

dcoombs-ocient: By this point, we shouldn't be throwing `ParseExecption` since we are done parsing.

I'd expect this function to throw `SQLException` and probably `IOException`.

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
```diff
@@ -0,0 +1,96 @@
...
+    // Extract the result set given its metadata and desired ExtractConfiguration.
+    public void extract(final ResultSet resultSet, final ResultSetMetaData resultSetMetaData) throws IOException, SQLException{
+
+        final int colCount = resultSetMetaData.getColumnCount();
+        // Grab the headers
+        final ArrayList<String> headers = new ArrayList<String>(colCount);
+        // Note that the columns of resultSetMetadata is 1 indexed.
+        for(int i = 1; i <= colCount; ++i){
+            headers.add(resultSetMetaData.getColumnName(i));
+        }
```
rveitch-Ocient, line 34: I was going to ask why we couldn't simply use an iterator from `ResultSetMetadata` but then glanced at the javadoc for this ancient API and wept.

rveitch-Ocient (commented):
In general LGTM, I'm not that familiar with the univocity parser but we'll be getting familiar with it soon for CSV support in LAT.

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
```diff
@@ -0,0 +1,96 @@
...
+        int currentRowInFile = 0;
+        try{
+            // next() can throw SQLException
+            while (resultSet.next()){
+                // The next call to getRow() will return a new row.
+                // This way, if the last row fits into the last line allowed by a file, then a new file will NOT be made.
+                if(maxRowsPerFile > 0 && currentRowInFile == maxRowsPerFile){
+                    // Hit the max on this current file. Create a new one.
+                    currentWriter.close();
```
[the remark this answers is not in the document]
tli-ocient: Yes it will throw an illegalstateexception. I think we want that to go all the way up. It will be indicative of a pretty critical error if it does show up though.

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
```diff
@@ -0,0 +1,85 @@
...
+        String fileName = resolveFileName(fileIndex);
+
+        // Construct our write pipeline. 
+        // Todo: Extract phase 2. Handle writing to S3. This means using S3OutputStream instead of FileOutputStream
+        OutputStream outputStream = null;
+        try {
+            outputStream = new FileOutputStream(fileName);
+            // Add a compression component if necessary.
+            if(extractConfig.getCompression() == ExtractConfiguration.Compression.GZIP){
+                outputStream = (OutputStream) new GZIPOutputStream(outputStream);
```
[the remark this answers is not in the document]
tli-ocient: took it out

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
tli-ocient: yes haha

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -36,7 +36,7 @@ public ExtractConfiguration(final Properties properties)
         } 
         catch (NullPointerException | IllegalArgumentException | ConversionException ex)
         {
-            LOGGER.log(Level.WARNING, "Caught exception in ExtractConfiguration constructor with message: %s", ex.getMessage());
+            LOGGER.log(Level.WARNING, String.format("Caught exception in ExtractConfiguration constructor with message: %s", ex.getMessage()));
```
dcoombs-ocient, line 39: Just thinking about this now. Since we ended up not using a logger for the CLI, should we just send this to stdout?

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
```diff
@@ -0,0 +1,96 @@
...
+        final int colCount = resultSetMetaData.getColumnCount();
+        // Grab the headers
+        final ArrayList<String> headers = new ArrayList<String>(colCount);
+        // Note that the columns of resultSetMetadata is 1 indexed.
+        for(int i = 1; i <= colCount; ++i){
+            headers.add(resultSetMetaData.getColumnName(i));
+        }
+        // The file count indexer. For File naming.
+        int fileIndex = 0; 
+        int maxRowsPerFile = extractConfig.getMaxRowsPerFile();
```
dcoombs-ocient, line 37: What does this default to? Its definitely an edge case, but I think we should make this an `Integer` and default it to `null` and skip the check entirely.

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
```diff
@@ -0,0 +1,96 @@
...
+            while (resultSet.next()){
+                // The next call to getRow() will return a new row.
+                // This way, if the last row fits into the last line allowed by a file, then a new file will NOT be made.
+                if(maxRowsPerFile > 0 && currentRowInFile == maxRowsPerFile){
+                    // Hit the max on this current file. Create a new one.
+                    currentWriter.close();
+                    currentWriter = recordExtractorFactory.create(fileIndex++);
+                    if(!extractConfig.getSkipHeader()){
+                        currentWriter.writeHeaders(headers);
+                    }
```
dcoombs-ocient, line 56: nit: This code is repeated, lets make it a function in case we need to add any more steps to starting a new file.

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
```diff
@@ -0,0 +1,96 @@
...
+                    // Reset the current row count.
+                    currentRowInFile = 0;
+                }
+                // Get the result row and write it. getRow() can throw SQLException.
+                Object[] row = getRow(resultSet, colCount);
+                currentWriter.writeRow(row);
+                currentRowInFile++;
+            }
+        } catch (SQLException ex) {
+            LOGGER.log(Level.WARNING, String.format("When extracting, caught a SQLException with message: %s", ex.getMessage()));
```
dcoombs-ocient, line 66: Where does this logger write to? Do we want it to write to the same log as JDBC? Or should we send this to stdout?

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: This will get sent to stdout via the exception thrown below. The CLI will catch and print the message.

dcoombs-ocient: Should we remove the log line then?

dcoombs-ocient: I asked for the log originally when I thought we were going to use a logger for the CLI.

dcoombs-ocient: Same goes for this line: https://github.com/Xeograph/ocient_jdbc/blob/d5b7cf60675f2309b6ea6a06c22fa38157c3a348/src/main/java/com/ocient/cli/extract/ExtractConfiguration.java#L239

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
dcoombs-ocient: Sorry, I missed the `maxRowsPerFile > 0` component of the check below. It would be better java style to use `Integer` and a null check instead, but that's a nit.

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: yea we could get rid of them.

tli-ocient: well, actually there may be some use for these in the future. How I understand these loggers work, we create (or retrieve) each logger like so:
`private static final Logger LOGGER = Logger.getLogger("com.ocient.cli.extract");`.
https://docs.oracle.com/javase/8/docs/api/java/util/logging/Logger.html#getLogger-java.lang.String-
 Then once we get the logger we can add a handler for them with addHandler:
https://docs.oracle.com/javase/8/docs/api/java/util/logging/Logger.html

So right now. These logs don't do anything because we haven't added a handler for them to be redirected to an output file. In the future, we could add such a thing and have all the logs that go to "com.ocient.cli.extract" be directed to a given log file. I think that could be useful.

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
tli-ocient: See comments above. It won't go to the JDBC logs.

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
tli-ocient: done

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
tli-ocient: changed

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
dcoombs-ocient: I don't see us adding a logger for the CLI until after the rewrite. I think anything we would want to see logged (Warnings and errors) we should write to stdout.

dcoombs-ocient (approved):

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
rveitch-Ocient: When logging is required, consider coding against the SLF4J facade rather than directly against java.util.Logging, which will provide for a richer feature set and more concise log statements. http://www.slf4j.org/

I think you could still use java.util.Logging as your backing implementation if you desire, though I'd recommend log4j2

dcoombs-ocient: I agree. Unfortunately the JDBC driver code already uses `java.util.Logging` directly, so Timmy was just being consistent here. We will fix that during our rewrite.

PR #201, opened 2021-11-16
[SOL-22] and [SOL-23] Implement CLI logging and exit code

tli-ocient:
https://ocient.atlassian.net/browse/SOL-22
https://ocient.atlassian.net/browse/SOL-23


src/main/java/com/ocient/cli/CLI.java
```diff
@@ -91,6 +93,7 @@ private static void connectTo(final String cmd)
 			{
 				System.out.println("Error: " + e.getMessage());
 				lastCommandErrored = true;
+				returnErrorCode = true;
```
dcoombs-ocient, line 96: Instead of setting this here, can we just set it in the `main` loop by checking the `lastCommandErrored` bool?

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1298,6 +1324,9 @@ else if (startsWithIgnoreCase(cmd, "SET TIMEOUT"))
 		{
 			setQueryTimeout(cmd);
 		}
+		else if(startsWithIgnoreCase(cmd, "SET CLI LOGGING")){
```
dcoombs-ocient, line 1327: Lets call this `CLI TRACE` since this isn't really logging (we are just writing to stdout). `CLI TRACE ON` and `CLI TRACE OFF`. It seems like most of these commands (besides timeout) leave out the `SET` portion.

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -937,7 +960,7 @@ public static void main(final String[] args)
 		}
 		catch (final UserInterruptException | EndOfFileException e)
 		{
-			return;
+			System.exit(returnErrorCode ? 1 : 0);
```
dcoombs-ocient, line 963: Can you make the `1` and `0` constants in the code. `EXIT_FAILURE` `EXIT_SUCCESS`, in case we add other exit calls in the future?

dcoombs-ocient (changes requested):

src/main/java/com/ocient/cli/CLI.java
tli-ocient: thats better. changed

src/main/java/com/ocient/cli/CLI.java
tli-ocient: ok changed.

src/main/java/com/ocient/cli/CLI.java
tli-ocient: yes, done.

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -416,7 +421,6 @@ private static void executeAndPrint(final String cmd)
 			{
 			}
 			System.out.println("Error: " + e.getMessage());
-			lastCommandErrored = true;
```
dcoombs-ocient, line 419: Why was this removed?

tli-ocient: My bad, accidentally cut when I was editing.

dcoombs-ocient (approved):

PR #203, opened 2021-11-18
User/tli/byte array wrapper and binary format

tli-ocient:
https://ocient.atlassian.net/browse/SOL-16
https://ocient.atlassian.net/browse/SOL-19


src/main/java/com/ocient/jdbc/XGByteArrayHelper.java
```diff
@@ -0,0 +1,18 @@
...
+    public static String bytesToHex(final byte[] bytes){
+        final char[] hexChars = new char[bytes.length * 2];
+		for (int j = 0; j < bytes.length; j++)
+		{
+			final int v = bytes[j] & 0xFF;
+			hexChars[j * 2] = hexArray[v >>> 4];
+			hexChars[j * 2 + 1] = hexArray[v & 0x0F];
+		}
+		return new String(hexChars);        
+    }
```
tli-ocient, line 15: This utility function was written before my time. I see other utilities that does bytes to hex conversion for us that we could just replace this with. I decided against it since this is used in the driver in a few places and I didn't want to change it in case there is some speed concerns.

src/main/java/com/ocient/cli/extract/ByteArrayWrapper.java
```diff
@@ -0,0 +1,35 @@
...
+        switch(binaryFormat){
+            case BASE64:{
+                return new String(Base64.encodeBase64(byteArray));
+            }
+            case UTF8: {
+                return new String(byteArray, StandardCharsets.UTF_8);
+            }
+            // Hexadecimal.
+            default: {
+                return  "0x" + XGByteArrayHelper.bytesToHex(byteArray);
```
tli-ocient, line 28: would it be better to case hexadecimal and default throw?

cj81499: A case for hex and default throw would be better imo. As is, if we add another `BinaryFormat` in the future, this code would (try to) use the hex format, rather than failing, which would direct a developer's attention to a place where additional logic is needed.

Depending on which version of Java this targets, we could use a [`switch` expression](https://docs.oracle.com/en/java/javase/17/language/switch-expressions.html) instead, and the compiler would check that all possible cases are covered.

The only other solution I can think of would be to make `ByteArrayWrapper` an interface, and use a factory to get an instance based on the format (ex: `Base64ByteArrayWrapper`). That way, the individual classes would only be responsible for logic specific to the associated `BinaryFormat`.

Given this class is as simple as it is today, a factory seems like overkill for now, but if you expect this class to become more complex, it could be a worthwhile choice. I don't know enough to make a prediction about how this code will change/grow over time, so I trust your judgement either way.

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
```diff
@@ -82,8 +82,7 @@ private CsvWriter startNewWriter(int fileIndex, final ArrayList<String> headers)
             if(resultSet.wasNull()){
                 element = extractConfig.getNullFormat();
             } else if (element instanceof byte[]){
```
cj81499, line 84: not sure what version of java we target, but potentially possible to use [pattern matching on `instanceof`](https://docs.oracle.com/en/java/javase/17/language/pattern-matching-instanceof-operator.html)

src/test/java/com/ocient/cli/extract/RecordExtractorFactoryTest.java
```diff
@@ -212,6 +215,7 @@ private XGResultSet makeFakeResultSet(int numRows){
             row.add(i);
             row.add(i + 5);
             row.add(null);
+            row.add(new String("deadbeef").getBytes(Charset.defaultCharset())); // This tests the byte array wrapper.
```
cj81499, line 218: no need for `new String`. `"deadbeef"` is already a `String`.

src/main/java/com/ocient/jdbc/XGTimestamp.java
```diff
@@ -272,12 +283,8 @@ public String toLocaleString()
...
-		final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
-		sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
-		final GregorianCalendar cal = new GregorianCalendar();
-		cal.setGregorianChange(new java.util.Date(Long.MIN_VALUE));
-		cal.setTimeZone(TimeZone.getTimeZone("UTC"));
-		sdf.setCalendar(cal);
-		return sdf.format(this) + "." + String.format("%09d", getNanos());
+		synchronized (sdf){
+			return sdf.format(this) + "." + String.format("%09d", getNanos());
+		}
```
tli-ocient, line 288: spotbugs called us out on this: https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#stcal-call-to-static-dateformat-stcal-invoke-on-static-date-format-instance
I added a synchronized keyword to get it to keep quiet. While moving this out into a static member does increase single threaded performance, if spotbugs is right and we have to synchronize on this, then it will limit our future multithread performance (since one thread can be in there at a time).

src/test/java/com/ocient/cli/extract/RecordExtractorFactoryTest.java
tli-ocient: changed

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
tli-ocient: i think our minimum version requirements for jdbc is java 11

src/main/java/com/ocient/cli/extract/ByteArrayWrapper.java
```diff
@@ -0,0 +1,35 @@
...
+            }
+        }
+    }
+
+    final private byte[] byteArray;
+    final private BinaryFormat binaryFormat;
```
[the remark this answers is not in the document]
tli-ocient: done

src/main/java/com/ocient/cli/extract/ByteArrayWrapper.java
tli-ocient: yea i agree, changed

src/main/java/com/ocient/cli/extract/ByteArrayWrapper.java
```diff
@@ -0,0 +1,39 @@
...
+    final private BinaryFormat binaryFormat;    
+
+    public ByteArrayWrapper(final byte[] byteArray, final BinaryFormat format){
+        this.byteArray = byteArray;
+        this.binaryFormat = format;
+    }
+
+    @Override
+    public String toString() {
+        switch(binaryFormat){
```
dcoombs-ocient, line 24: This may be a premature optimization, but since the `binaryFormat` is known at the time where we create the `ByteArrayWrapper` we could make a `ByteArrayWrapperFactory` that does this switch at construction and not at runtime.

I think this is okay to leave for now. We will probably get some branch prediction that will make the performance impact negligible.

src/main/java/com/ocient/cli/extract/ByteArrayWrapper.java
```diff
@@ -0,0 +1,39 @@
...
+                return new String(Base64.encodeBase64(byteArray), Charset.defaultCharset());
+            }
+            case UTF8: {
+                return new String(byteArray, StandardCharsets.UTF_8);
+            }
+            case HEXADECIMAL:{
+                return  "0x" + XGByteArrayHelper.bytesToHex(byteArray);
+            }
+            default: {
+                throw new IllegalArgumentException(String.format("Invalid binary format specified: %s", binaryFormat.toString()));
```
dcoombs-ocient, line 35: Can you do this check in the constructor? I'd like to know that there is an invalid binary format at construction. 

You'll probably have to leave this check here to make the compiler happy, but we should catch this in the constructor.

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
dcoombs-ocient: We will be on Java 11 for the foreseeable future since this is a client-side jar (not everyone is up to speed on their java versions).

src/main/java/com/ocient/jdbc/XGByteArrayHelper.java
dcoombs-ocient: Thats fine with me.

src/main/java/com/ocient/jdbc/XGTimestamp.java
dcoombs-ocient: Well that's annoying. This might be why the code wasn't written like this in the first place.

What about using `ThreadLocal<T>` https://stackoverflow.com/questions/16249687/java-threadlocal-static?

src/main/java/com/ocient/cli/extract/ByteArrayWrapper.java
cj81499: I think you're discussing the same thing I brought up here: https://github.com/Xeograph/ocient_jdbc/pull/203#discussion_r754672365

src/main/java/com/ocient/cli/extract/ByteArrayWrapper.java
cj81499: imo, this is probably the strongest reason to refactor to a use a `ByteArrayWrapperFactory` now, rather than waiting.

src/main/java/com/ocient/jdbc/XGTimestamp.java
tli-ocient: This is pretty neat. I was worried we'd have to implement something like this. I moved construction of the SDF back to inside of toString, since we want each thread to execute it once.

src/main/java/com/ocient/jdbc/XGTime.java
```diff
@@ -267,12 +269,17 @@ public String toLocaleString()
...
+			// This is the first time this thread has constructed this class (XGTime).
+			SimpleDateFormat newSDF = new SimpleDateFormat("HH:mm:ss");
+			newSDF.setTimeZone(TimeZone.getTimeZone("UTC"));
+			final GregorianCalendar cal = new GregorianCalendar();
+			cal.setGregorianChange(new java.util.Date(Long.MIN_VALUE));
+			cal.setTimeZone(TimeZone.getTimeZone("UTC"));
+			newSDF.setCalendar(cal);
+			threadLocalSDF.set(newSDF);
+		}		
+		return threadLocalSDF.get().format(this) + "." + String.format("%09d", getNanos());
```
cj81499, line 283: I don't think this logic is quite correct because it's possible that multiple threads could get `null`, create `sdf`, and call `set` on `threadLocalSDF`.

I don't have any prior experience with `ThreadLocal` in Java, but the [docs](https://docs.oracle.com/javase/7/docs/api/java/lang/ThreadLocal.html) make it sound like the intended approach is to extend it, overrideing the `initialValue` method to construct and return the value you want to be thread local.

cj81499 (changes requested):

src/main/java/com/ocient/jdbc/XGTime.java
tli-ocient: I'm not familiar with this class either. 
` it's possible that multiple threads could get null, create sdf, and call set on threadLocalSDF`
My impression was that this was a map with the read as a key to the value. Thus the multiple threads would be getting "different" nulls and then setting `threadLocalSDF` for different keys.

Regardless, I was doing something else wrong. We cannot initialize the initial value to be null like I was doing or else it would throw a null exception. We don't need to do that anyway. We can just set it to be exactly what want by overriding initial value like you said. I think this is correct now.

dcoombs-ocient: Looks good now.

dcoombs-ocient (approved):

src/main/java/com/ocient/jdbc/XGTime.java
```diff
@@ -14,6 +14,18 @@
 	/** */
 	private static final long serialVersionUID = -6636929210798184587L;
 
+	private static ThreadLocal<SimpleDateFormat> threadLocalSDF = new ThreadLocal<SimpleDateFormat>(){
```
cj81499, line 17: Since this and the `ThreadLocal` in `XGTimestamp.java` are identical, except for the date format string, do you think it's worth refactoring into a class that takes a date format string parameter?

cj81499 (approved):
One last concern, otherwise LGTM.

src/main/java/com/ocient/jdbc/XGTime.java
tli-ocient: We might even want to refactor some of the other toString methods in these classes (There's more than just toString), but I don't want to do that right now

PR #204, opened 2021-11-18
User/tli/more extract config

tli-ocient:
https://ocient.atlassian.net/browse/SOL-19


src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -234,4 +271,21 @@ private void validateConfiguration() throws ParseException
             System.out.println(String.format("Location type was not S3 (was %s) and a bucket (%s) was specified. Bucket will be ignored.", locationType.name(), bucket));
         }
     }
+
+    // There doesn't seem to be a getChar helper in ImmutableConfiguration. So we take it as a string and 
+    // use the first char.
+    private char getCharFromConfigHelper(ImmutableConfiguration config, final String key, final char defaultChar) throws ParseException
```
tli-ocient, line 277: Definitely open to suggestions on how to do this better if possible. The problem is there isn't a getChar from ImmutableConfiguration (Configuration). So I rely on taking a 1 length string.

tli-ocient: Univocity parser takes chars for quote and quoteEscape options

cj81499: The generic get method might work? If not, this seems fine to me.

https://commons.apache.org/proper/commons-configuration/apidocs/org/apache/commons/configuration2/ImmutableConfiguration.html#get-java.lang.Class-java.lang.String-T-

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -31,6 +33,8 @@ public ExtractConfiguration(final Properties properties)
             compression = Compression.valueOf(config.getString(COMPRESSION, DEFAULT_COMPRESSION).toUpperCase());            
             maxRowsPerFile = config.getInteger(MAX_ROWS_PER_FILE, DEFAULT_MAX_ROWS_PER_FILE);
             skipHeader = config.getBoolean(SKIP_HEADER, DEFAULT_SKIP_HEADER);
+            // This Charset.forName will throw illegalArgumentException if name is not valid.
+            encoding = (config.getString(ENCODING) != null) ? Charset.forName(config.getString(ENCODING)) : DEFAULT_ENCODING;
```
cj81499, line 37: Might be clearer to use `Optional`? I mostly like this because it avoids calling `config.getString(ENCODING)` multiple times.

```suggestion
            encoding = Optional.ofNullable(config.getString(ENCODING)).map(Charset::forName).orElse(DEFAULT_ENCODING);
```

tli-ocient: ok, changed

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: this probably won't do it, since I'd have to use `String` as the first argument, since that is what it is when it gets entered by the syntax parser.

cj81499: I tried the following, and it worked how I expected (how the comments say):


```java
import java.util.Map;
import org.apache.commons.configuration2.MapConfiguration;

class Playground {
  public static void main(final String[] args) {
    final var mc = new MapConfiguration(Map.ofEntries(Map.entry("1", "!"), Map.entry("2", "oops")));

    // prints "!"
    final var c1 = mc.get(Character.class, "1");
    System.out.println(c1);

    // throws org.apache.commons.configuration2.ex.ConversionException
    final var c2 = mc.get(Character.class, "2");
    System.out.println(c2);
  }
}
```

tli-ocient: thanks, that just works

dcoombs-ocient (approved):

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -134,6 +150,12 @@ public ExtractConfiguration(final Properties properties)
     private final boolean skipHeader;
     // Format string to use for writing NULL values to the output files.
     private final String nullFormat;
+    // Encoding to use when writing out bytes.
+    private final Charset encoding;
+    // Character used to escape quotes.
+    private final Character escape;
```
cj81499, line 156: this can still be a `char`. Java will auto unbox the `Character` for us. https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html

cj81499 (changes requested):

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: Yea I noticed this in `RecordExtractorFactory.java` where I didn't need to convert it myself. This is pretty neat. I am going to keep it this way since Dan had said it was more Java like to use Integer rather than int for max rows.

dcoombs-ocient: I said that in the use case where the configuration variable could be `null`. I don't think this one can ever be `null`? In that case its better to use a `char` over a `Character`.

tli-ocient: Ah ok, I'll switch this.

tli-ocient: https://github.com/Xeograph/ocient_jdbc/pull/207

PR #205, opened 2021-11-20
first extract syntax parser

tli-ocient:
https://ocient.atlassian.net/browse/SOL-11
A lot of this code is from:
https://github.com/Xeograph/ocient_jdbc/pull/199

Please read:
This will generally work. We need to quote any values that contain any alphanumeric characters with double quotes.I know this isn't perfect. There are two weaknesses here.

1. `)`  cannot be used as a config value. This is because of how we first attempt to match `OPTIONS(...)`.  If this does become a problem, we can try to address with a more complicated pattern in the first regex pass.
2. We need to reserve `"` to quote our arguments. This means that `"` cannot be used as a config value either. I could not put together a regex that allows me to split strings allowing for quoted segments and escaped quotes within the quoted segments. We need to get quote escape working. Even if I could, the cli has this `scrubCommand` which gets called on user input before anything else. This preceeds all commands that go into the CLI and does not currently handle escaped quotes. We would need to fix this as well.

I don't see 1 being a big deal. For 2, there is only 1 argument where it would be reasonable to use `"` as an argument, `FIELD_OPTIONALLY_ENCLOSED_BY` which is used to surround quoted strings. But `"` is the default there, so that should be fine as well.

I do expect that these short comings, especially 2 , will need to be readdressed. We need to come up with a strategy for this. 
Maybe get some help from people who know regex better, implement a state machine parser, bring in something like `antlr` or something else. For now though, I think these short comings are "ok" for a phase 1 merge and demo. I would like to get something working to Selva by Wenesday of next week so that he can test and be ready for the demo.


src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1441,6 +1442,7 @@ public static void extractTo(final String cmd)
 		catch (ParseException e)
 		{
 			System.out.println(String.format("Extract syntax parsing failed with: %s", e.getMessage()));
+			lastCommandErrored = true;
```
cj81499, line 1445: rather than remembering to add this to each catch branch, it might be easier to "assume" an error (by setting `lastCommandErrored = true;`) prior to the `try/catch`, and set it to `false` if we make it past the `try/catch`.

tli-ocient: great point - changed

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
```diff
@@ -0,0 +1,132 @@
...
+    private static String QUERY_GROUP = "query";
+    private static String ARG_GROUP = "arg";
+    private static String VAL_GROUP = "val";
+    private static String QUOTED_GROUP = "quoted";
+    private static String UNQUOTED_GROUP = "unquoted";
+
+    // Courtesy of https://stackoverflow.com/questions/1757065/java-splitting-a-comma-separated-string-but-ignoring-commas-in-quotes?rq=1
+    private static String TOKEN_SPLITTER_STRING = new String(",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)");
+
+    private static final Pattern EXTRACT_PATTERN = Pattern.compile(
```
dcoombs-ocient, line 28: I'm going to trust that this works, although it may be worth adding a bunch more tests for all of the weird cases.

Can you write something up on the design doc that explains the limitations of this parsing? You can put it under this section https://ocient.atlassian.net/wiki/spaces/PM/pages/3173679251/Design+-+Extract+Data+to+Delimited+Files#Syntax-Parsing and title Phase 1 Syntax Parsing Deficiencies.

dcoombs-ocient: I'd prefer if we did something like this in user docs, but since the userdocs live on the xgsrc side we have to update them when we merge the feature branch in. For now we will basically use the design doc as our userdocs.

dcoombs-ocient (approved):

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
tli-ocient: I updated the design doc with my comment at the top of this PR.

PR #208, opened 2021-11-30
[SOL-17] User/tli/s3 output stream

tli-ocient:
https://ocient.atlassian.net/browse/SOL-17

Tests still need to be added


src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -102,6 +105,12 @@ public ExtractConfiguration(final Properties properties)
     public static final String BUCKET = "bucket";
     public static final String DEFAULT_BUCKET = null;
 
+    public static final String REGION = "region";
+    public static final String DEFAULT_REGION = null;
```
tli-ocient, line 109: Should region have a default?

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
```diff
@@ -80,6 +92,31 @@ private void tryCloseOutputStream(OutputStream outputStream){
...
+        String awsAccessKey = defaultCredentials.getAWSAccessKeyId();
+        String awsSecretKey = defaultCredentials.getAWSSecretKey();
+        // If either awsAccessKey or awsSecretKey is explicitly specified, use those instead.
+        if(!extractConfig.getAwsKeyId().equals("")){
+            awsAccessKey = extractConfig.getAwsKeyId();
+        }
+
+        if(!extractConfig.getAwsKeySecret().equals("")){
+            awsSecretKey = extractConfig.getAwsKeySecret();
+        }
```
tli-ocient, line 106: Should we require that both be specified or neither?

src/main/java/com/ocient/cli/extract/S3OutputStream.java
```diff
@@ -0,0 +1,149 @@
...
+        assertOpen();
+        if(bufferPosition >= buffer.length){
+            flush();
+        }
+        buffer[bufferPosition++] = (byte) b;
+    }
+
+    // Flushes the output stream. This generates a new part.
+    @Override
+    public synchronized void flush(){
```
tli-ocient, line 103: There is a rule stated here: https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html
that each part, except the last, needs to be 5MB.
This isn't a problem now since we are the only ones using this class. However, if we were to give it to someone else, there could be a situation where they open this stream, write a tiny amount, flush, write another tiny amount, and flush again.

This will break that rule and result in an exception being thrown stating
`Your proposed upload is smaller than the minimum allowed size`.

We could change this `flush` to do nothing, and add a `flushInternal` method which does what this does. Then call `flushInternal` everywhere we need it in this file. Of course then `flush` would not actually synchronize the bytes when they call it. Thoughts?

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
dcoombs-ocient: In the LAT we defaulted to "us-east-2" https://ocient.gitlab.io/lat/userguide.html#_s3_source_specific_configuration

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
dcoombs-ocient: Yeah that sound like a good lint.

src/main/java/com/ocient/cli/extract/S3OutputStream.java
dcoombs-ocient: I think we should have this `flush` do nothing and clearly state that in the comment. Basically, for this output stream, nobody should be calling flush and should only use `close()` when complete.

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1475,7 +1476,7 @@ public static void extractTo(final String cmd)
 		{
 			rsExtractor.extract(resultSet, rsMetaData);
 		} 
-		catch (IllegalStateException | IllegalArgumentException | IOException | SQLException e)
+		catch (IllegalStateException | IllegalArgumentException | IOException | SQLException | SdkClientException e)
```
dcoombs-ocient, line 1479: I don't think our interface should leak the `SdkClientException`, you should catch that somewhere else and wrap it into one of the other exceptions here, probably `IOException`.

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
```diff
@@ -80,6 +92,31 @@ private void tryCloseOutputStream(OutputStream outputStream){
...
+        // Create the credentials
+        final BasicAWSCredentials basicAWSCredentials = new BasicAWSCredentials(awsAccessKey, awsSecretKey);
```
dcoombs-ocient, line 108: What about for Anonymous access (when there are no keys specified)?

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
```diff
@@ -80,6 +92,31 @@ private void tryCloseOutputStream(OutputStream outputStream){
...
+
+        AmazonS3ClientBuilder builder = AmazonS3ClientBuilder.standard()
+            .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(extractConfig.getEndpoint(), extractConfig.getRegion()))
+            .withCredentials(new AWSStaticCredentialsProvider(basicAWSCredentials));
+        builder.setPathStyleAccessEnabled(extractConfig.getPathStyleAccess());
```
dcoombs-ocient, line 113: Any reason this is split into its own line statement? Can it just be added to the previous?

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
```diff
@@ -80,6 +92,31 @@ private void tryCloseOutputStream(OutputStream outputStream){
...
+        // Build the client.
+        AmazonS3 s3Client = builder.build();
```
dcoombs-ocient, line 115: Same here.

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
```diff
@@ -35,7 +41,13 @@ CsvWriter create(int fileIndex) throws IOException{
         // Todo: Extract phase 2. Handle writing to S3. This means using S3OutputStream instead of FileOutputStream
         OutputStream outputStream = null;
         try {
-            outputStream = new FileOutputStream(fileName);
+            if(extractConfig.getLocationType() == ExtractConfiguration.LocationType.S3){
+                // S3
+                outputStream = makeS3OutputStream(fileName);
```
dcoombs-ocient, line 46: Just noting this so we don't forget, this may need to be proved out in testing.

I don't think we want to wrap this `outputStream` in a `BufferedWriter` (like you do below), because the `S3OutpoutStream` already buffers. We should make a ticket to test this out. If the performance is better (or the same, since it uses less memory) we should remove the `BufferedWriter`.

src/main/java/com/ocient/cli/extract/S3OutputStream.java
```diff
@@ -0,0 +1,149 @@
...
+        // Start the upload.
+        final InitiateMultipartUploadRequest request = new InitiateMultipartUploadRequest(bucket, filePath);
+        InitiateMultipartUploadResult initResponse = s3Client.initiateMultipartUpload(request);
+        // Store the upload id. This is the id for the entire upload.
+        this.uploadId = initResponse.getUploadId(); 
+    }
+
+    // Write a byte array to the S3OutputStream
+    @Override
+    public void write(byte[] byteArray){
```
dcoombs-ocient, line 64: TIL: https://docs.oracle.com/javase/7/docs/api/java/io/OutputStream.html#write(byte[])

Interesting they would design their API like that. I would have expected it to act like a stream and append rows to the next offset...

src/main/java/com/ocient/cli/extract/S3OutputStream.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+        // Now upload the part.
+        uploadPart();
+        // Reset the position.
+        bufferPosition = 0;
+    }
+
+    // Upload a part and store the eTag.
+    protected void uploadPart(){
+        UploadPartResult uploadResult = s3Client.uploadPart(new UploadPartRequest()
```
dcoombs-ocient, line 117: What exceptions does this throw? We probably want to catch them and adhere to the `write()` interface contract and re-throw an `IOException`

src/main/java/com/ocient/cli/extract/S3OutputStream.java
```diff
@@ -0,0 +1,149 @@
...
+import com.amazonaws.services.s3.model.PartETag;
+import com.amazonaws.services.s3.model.UploadPartRequest;
+import com.amazonaws.services.s3.model.UploadPartResult;
+
+import java.io.ByteArrayInputStream;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.List;
+
+public class S3OutputStream extends OutputStream{
```
dcoombs-ocient, line 17: The logic here LGTM. I think we should try and get some tests written for this class using a mock `s3Client`.

@rveitch-Ocient @cj81499 @ejiang-Ocient @zzpxyx, I'm not super well versed in Java mocks, can one of you point Timmy to examples in the LAT codebase?

dcoombs-ocient (changes requested):
Mostly LGTM, a few small comments. I'd like to get some tests added if possible.

src/main/java/com/ocient/cli/extract/S3OutputStream.java
zzpxyx: In LAT, we are using JUnit plus Mockito. We just mock an S3 client as any other mocked objects: https://gitlab.com/ocient/lat/-/blob/main/src/test/java/com/ocient/lat/source/file/s3/S3FileFetcherTest.java#L44.

pom.xml
```diff
@@ -131,8 +131,13 @@
         <groupId>commons-codec</groupId>
         <artifactId>commons-codec</artifactId>
         <version>1.15</version>
-    </dependency>        
-  </dependencies>
+    </dependency>
+    <dependency>
+        <groupId>com.amazonaws</groupId>
+        <artifactId>aws-java-sdk-s3</artifactId>
+        <version>1.12.119</version>
```
zzpxyx, line 138: Just curious, any reason to use the V1 AWS Java SDK as opposed to [V2](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/home.html)?

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
tli-ocient: combined

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
tli-ocient: combined, I missed that there was a `withPathStyleAccessEnabled` as well. Was using `setPathStyleAccessEnabled` which returns void and could not chain.

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
tli-ocient: I am not sure what you mean. When neither keys are specified, the default provider chain is used above to find them

tli-ocient: well I should say, they are always found by the default provider chain and overwritten if the user specifies anything. I should probably change this to only look with the default provider chain if the user doesn't specify.

dcoombs-ocient: What do you mean "always found by the default provider chain"? What does it do if it can't find credentials anywhere? https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html

tli-ocient: Sorry not "always found". If it is not found then, looking at their code, an exception is thrown.
`
        throw new SdkClientException("Unable to load AWS credentials from any provider in the chain: "
                                     + exceptionMessages);
`

dcoombs-ocient: This is what I mean by "anonymous access" not all S3 implementations need keys (COS in riverside for example should be setup for "anonymous access").

tli-ocient: To summarize, they need to be able to write to the server even if the no credentials are found if the server allows for anonymous access?

dcoombs-ocient: Correct

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: region now defaults to "us-east-2"

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
tli-ocient: ExtractConfiguration now requires that both are specified or neither. This now proceeds with that assumption

src/main/java/com/ocient/cli/extract/S3OutputStream.java
rveitch-Ocient: In addition to `mock()`, `spy()` is also good to know about for when you want to use a real implementation but override a particular method for your test. Useful when the test orchestration is complex and you'd rather drive most of it with the real code rather than mocking a complex set of behaviors to mirror what the real code would do.

rveitch-Ocient: mockito docs: https://site.mockito.org/

src/main/java/com/ocient/cli/extract/S3OutputStream.java
tli-ocient: moved logic out into a `flushInternal` method which is now used. `flush` now does nothing.

src/main/java/com/ocient/cli/extract/S3OutputStream.java
tli-ocient: thanks, I will take a look at this and add some mock tests

pom.xml
tli-ocient: Oh I didn't realize there was a V2.
Is the proper dependency for V2
`    <dependency>
      <groupId>software.amazon.awssdk</groupId>
      <artifactId>s3</artifactId>
      <version>2.17.53</version>
    </dependency>`

I copied this from the LAT project

dcoombs-ocient: Yes although you might as well get onto `2.17.93` as it is the latest: https://mvnrepository.com/artifact/software.amazon.awssdk/s3

zzpxyx: According to the V2 doc, it's `2.17.93` at the moment: https://github.com/aws/aws-sdk-java-v2#using-the-sdk. Note that V1 and V2 may have some slight difference in API naming, packaging, and usage. If we are in a hurry, using V1 is OK-ish since it's not fully deprecated. If we have time, I'd recommend converting it to the future-proofing V2.

tli-ocient: We have time. Theres not too much to convert, I'll do this now

src/main/java/com/ocient/cli/CLI.java
tli-ocient: wrapped it all instances that throw SdkClientException in IOException for `S3OutputStream.java`

src/main/java/com/ocient/cli/extract/S3OutputStream.java
tli-ocient: wrapped

pom.xml
tli-ocient: Changed to using V2. Most of it is pretty straightforward, though I do have one question for @zzpxyx . Do you know what `DefaultCredentialsProvider.create()` does if it cannot find the default credentials set up anywhere? The equivalent of this in V1 throws an exception, but this looks it doesn't. So how does this behave if no credentials is not set up anywhere in this list:
https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/auth/credentials/DefaultCredentialsProvider.html

zzpxyx: Looking at the source code, I think it will throw an exception, no? https://github.com/aws/aws-sdk-java-v2/blob/4289e9a98c8db6b25643c40c9e77fdae6af1d5d5/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/AwsCredentialsProviderChain.java#L109

tli-ocient: Ok, thanks, I missed that. This should all be fine then.

src/main/java/com/ocient/cli/extract/S3OutputStream.java
tli-ocient: added mock tests

PR #212, opened 2021-12-02
try fixing set socket options again

tli-ocient:
we missed a few things


src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1600,12 +1600,16 @@ private void tryConfigureSocketOptions(Socket sock){
 				sock.setOption(entry.getKey(), entry.getValue());
 			} catch (IOException | UnsupportedOperationException | IllegalArgumentException | SecurityException ex) {
 				LOGGER.log(Level.WARNING, String.format("Failed setting socket option with message %s."), ex.getMessage());
+			} catch (NoSuchMethodError error){
```
jasonar81, line 1603: I thought that the exception that was thrown was not one of these but was like no such field or something.

jasonar81: I think we really could just catch(Exception e) here

tli-ocient: catching error won't catch exception: https://stackoverflow.com/questions/38283476/java-nosuchmethoderror-not-caught-by-exception. But I could change the above to catch exception instead of all of those other things.

tli-ocient: Yes I changed it to just catch all Exceptions in the above line. I don't want to end up in another situation with an exception we didn't cover.

jasonar81 (approved):

PR #213, opened 2021-12-06
[SOL-18] multiThreadedResultSetExtractor

tli-ocient:
https://ocient.atlassian.net/browse/SOL-18


src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1471,6 +1462,22 @@ public static void extractTo(final String cmd)
...
+		ResultSetExtractor rsExtractor = null;
+		try
+		{	
+			// Build the extractor
+			ExtractConfiguration config = new ExtractConfiguration(parseResult.getConfig());
+			if(config.getAllowMultiThreading() && ((XGResultSet) resultSet).getNumClientThreads() > 1){
+				rsExtractor = new MultiThreadedResultSetExtractor(config);
+			} else {
+				rsExtractor = new SingleThreadedResultSetExtractor(config);
+			}
```
cj81499, line 1474: nit, this might be more readable as a ternary and by giving the complicated condition a meaningful variable name.

```java
boolean useMultiThreadedResultSetExtractor = config.getAllowMultiThreading() && ((XGResultSet) resultSet).getNumClientThreads() > 1;
rsExtractor = useMultiThreadedResultSetExtractor ? new MultiThreadedResultSetExtractor(config) : new SingleThreadedResultSetExtractor(config);
```

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -265,6 +271,11 @@ public boolean getPathStyleAccess()
         return pathStyleAccess;
     }
 
+    public boolean getAllowMultiThreading()
```
cj81499, line 274: java style nit. For boolean getters, I prefer using `is` as a prefix, over `get` because it typically reads more closely to english. (ex: `if (animal.isHeavy()) { ... }` vs `if (animal.getHeavy()) { ... }`)

I'd call this `isMultithreadingAllowed`. I'd also probably change the property to `multithreadingAllowed` to match, but I'm less concerned with that, since it's private.

src/main/java/com/ocient/cli/extract/MultiThreadedResultSetExtractor.java
```diff
@@ -0,0 +1,145 @@
...
+    // Synchronized thread printing helper.
+    private synchronized void printError(String message){
+        System.out.println(message);
+    }
+    // flag for signaling threads to join
+    private final AtomicBoolean isFinishedAdding;
+    // Result queues for each thread
+    ArrayList<LinkedBlockingQueue<Object[]>> resultQueues;
+    // Writing threads
+    ArrayList<Thread> writerThreads;
```
cj81499, line 144: should these be `private`?

Also, why do we declare fields at the bottom of the `class`? I've seen them at the start of the `class` in every other Java project I've worked on.

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
```diff
@@ -81,8 +83,13 @@ CsvWriter create(int fileIndex) throws IOException{
...
     // Simple utility function to help resolve the file name.
-    private String resolveFileName(int fileIndex){
-        String newFileName = extractConfig.getFilePrefix() + String.valueOf(fileIndex) + extractConfig.getFileExtension();
+    private String resolveFileName(int fileIndex, int threadNumber){
+        String newFileName = null;
+        if(isMultiThreaded){
+            newFileName = extractConfig.getFilePrefix() + String.valueOf(threadNumber) + "_" + String.valueOf(fileIndex) + extractConfig.getFileExtension();
+        } else {
+            newFileName = extractConfig.getFilePrefix() + String.valueOf(fileIndex) + extractConfig.getFileExtension();
+        }
```
cj81499, line 92: might be more readable?

```suggestion
        String newFileName = extractConfig.getFilePrefix();
        if (isMultiThreaded) {
            newFileName += String.valueOf(threadNumber) + "_";
        }
        newFileName += String.valueOf(fileIndex) + extractConfig.getFileExtension();
```

src/main/java/com/ocient/cli/extract/MultiThreadedResultSetExtractor.java
```diff
@@ -0,0 +1,145 @@
...
+        } catch (SQLException ex){
+            // Need to fail and cleanup.
+            joinThreads();
+            printError(String.format("Main thread encountered error while attempting to fetch row from result set with message: %s", ex.getMessage()));
+        }
+        // Join the threads.
+        joinThreads();
+    }
+
+    private void joinThreads(){
```
cj81499, line 68: Is it safe to potentially call this multiple times, or do we need to return within the `catch` blocks?

pom.xml
```diff
@@ -139,10 +139,10 @@
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
+      <artifactId>mockito-inline</artifactId>
```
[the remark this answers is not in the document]
tli-ocient: I needed this to mock XGResultSet which is a final class.
https://stackoverflow.com/questions/14292863/how-to-mock-a-final-class-with-mockito

If theres another way to do this, then I am happy to change this. I don't want to make XGResultSet a non final class though.

src/main/java/com/ocient/cli/extract/ResultSetExtractor.java
```diff
@@ -9,61 +9,25 @@
...
+    ResultSetExtractor(final ExtractConfiguration config, boolean isMultiThreaded){
         extractConfig = config;
-        recordExtractorFactory = new RecordExtractorFactory(config);
+        recordExtractorFactory = new RecordExtractorFactory(config, isMultiThreaded);
+        headers = new ArrayList<String>();
     }
 
-    // Extract the result set given its metadata and desired ExtractConfiguration.
-    public void extract(final ResultSet resultSet, final ResultSetMetaData resultSetMetaData) throws IllegalStateException, IOException, SQLException{
+    // Should be overrided by inheriting class
```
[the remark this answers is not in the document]
tli-ocient: removed comment

src/main/java/com/ocient/cli/extract/MultiThreadedResultSetExtractor.java
tli-ocient: good catch. forgot to put the explicit returns there

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
tli-ocient: changed

src/main/java/com/ocient/cli/extract/MultiThreadedResultSetExtractor.java
tli-ocient: changed and moved members to the top of class

src/main/java/com/ocient/cli/extract/MultiThreadedResultSetExtractor.java
```diff
@@ -0,0 +1,145 @@
...
+                thread.join();            
+            } catch (InterruptedException ex){
+                // Shouldn't ever happen because this thread is the only one capable of interrupting these writer threads.
+                printError(String.format("Joined thread was interrupted with message: %s", ex.getMessage()));
+            }
+        }   
+    }
+
+    // Implementation of writing threads
+    private class WriterThread implements Runnable{
```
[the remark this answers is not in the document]
tli-ocient: changed to extend `Thread`

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: changed

src/main/java/com/ocient/cli/CLI.java
tli-ocient: changed

src/main/java/com/ocient/cli/extract/RecordExtractorFactory.java
```diff
@@ -63,8 +70,11 @@ CsvWriter create(int fileIndex) throws IOException{
             throw ex;
         }
         // Generate the output stream writer with the proper encoding.
-        OutputStreamWriter outputStreamWriter = new OutputStreamWriter(outputStream, extractConfig.getEncoding());
-        BufferedWriter bufferedWriter = new BufferedWriter(outputStreamWriter);
+        Writer writer = new OutputStreamWriter(outputStream, extractConfig.getEncoding());
+        // If we are extracting locally, then add a buffered writer.
+        if(extractConfig.getLocationType() == ExtractConfiguration.LocationType.LOCAL){
+            writer = new BufferedWriter(writer);
+        }
```
tli-ocient, line 77: I just threw this in this PR too. After some testing, I don't see having this buffered writer making a difference when we are uploading since the S3OutputStream writer already has buffering.

pom.xml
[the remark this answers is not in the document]
dcoombs-ocient: I think you should use what they recommend here: https://javadoc.io/static/org.mockito/mockito-core/3.7.7/org/mockito/Mockito.html#39

```
This mock maker is turned off by default because it is based on completely different mocking mechanism that requires more feedback from the community. It can be activated explicitly by the mockito extension mechanism, just create in the classpath a file /mockito-extensions/org.mockito.plugins.MockMaker containing the value mock-maker-inline. 
```

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1471,6 +1462,21 @@ public static void extractTo(final String cmd)
...
+		ResultSetExtractor rsExtractor = null;
+		try
+		{	
+			// Build the config
+			ExtractConfiguration config = new ExtractConfiguration(parseResult.getConfig());
+			// Build the extractor
+			boolean useMultiThreadedResultSetExtractor = config.isMultiThreadingAllowed() && ((XGResultSet) resultSet).getNumClientThreads() > 1;
+			rsExtractor = useMultiThreadedResultSetExtractor ? new MultiThreadedResultSetExtractor(config) : new SingleThreadedResultSetExtractor(config);
+			
+		}
```
dcoombs-ocient, line 1474: I prefer this parsing occurs before the query is submitted, such that we don't submit a query to the database and need to cancel it if parsing fails.

How much of this can we push back above running the query?

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -130,6 +131,9 @@ public ExtractConfiguration(final Properties properties)
     public static final String PATH_STYLE_ACCESS = "path_style_access";
     public static final boolean DEFAULT_PATH_STYLE_ACCESS = false;
 
+    public static final String MULTITHREADING_ALLOWED = "allow_multithreading";
+    public static final boolean DEFAULT_MULTITHREADING_ALLOWED = false;
```
dcoombs-ocient, line 135: We should discuss with product if we want this to be on or off by default. I believe our JDBC multi-threading is on by default, correct?

src/main/java/com/ocient/cli/CLI.java
tli-ocient: We can only push the construction of the `ExtractConfiguration` back up. In order to know the number of threads used, we need to have executed the query and have the result set.

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
tli-ocient: Correct JDBC multi-threading is on by default. My thought in having this off by default is the user may be annoyed by having their result file split into multiple pieces unless they explicitly signed up for it.

dcoombs-ocient (commented):
The logic here looks good to me! One comment on trying to parse/build as much as possible before the query starts running.

We should insure we do some non-happy-path testing with multi-threading on to ensure that everything correctly cleans up/shuts down when a query fails for whatever reason.

src/main/java/com/ocient/cli/CLI.java
dcoombs-ocient: What can fail in the constructors for the `ResultSetExtractor`'s? Is there anything we can lint in the `ExtractConfiguration` or through another method such that we minimize the changes of anything failing after the query is submitted?

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
dcoombs-ocient: That makes sense to me.

src/main/java/com/ocient/cli/CLI.java
tli-ocient: I think we already do as much as we can with regards to linting in `ExtractConfiguration`. Also submitting the query does not mean we wait for it to finish running. We don't block waiting for all the results to return. So it's not as much time lost as it initially may seem.

pom.xml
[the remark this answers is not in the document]
rveitch-Ocient: I experimented with mocking static methods in the past (not final classes) using the inline mockito stuff, and one drawback is that it mucks with the bytecode in a way that makes stepping through the source in a debugger somewhat confusing, you'll sometimes skip past source lines you'd think you should be able to reach but are unable to. Just a heads up, may not be a dealbreaker here.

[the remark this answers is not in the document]
tli-ocient: Rob and I were talking about this for a bit ^. It doesn't seem to be a deal breaker since this is in strictly test code, but I thought we'd make you aware.

src/main/java/com/ocient/cli/CLI.java
tli-ocient: moved `ExtractConfiguration` construction up.

pom.xml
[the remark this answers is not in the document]
tli-ocient: Changed to alternate way to set up mockito as suggested by https://javadoc.io/static/org.mockito/mockito-core/3.7.7/org/mockito/Mockito.html#39

dcoombs-ocient (approved):

PR #215, opened 2021-12-10
[SOL-25] User/tli/extract syntax improvements

tli-ocient:
https://ocient.atlassian.net/browse/DB-18851
https://ocient.atlassian.net/browse/SOL-25


src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
```diff
@@ -129,4 +136,72 @@ private static Properties parseOptions(String ops) throws ParseException
...
+     * Ex: key1=val1,key2="val2",key3=val3 -> [key1=val1, key2=val2, key3=val3]
+     * @param options option string containing all the options.
+     * @return ArrayList<String> seperated key=value pairs.
+     * @throws ParseException
+     */
+
+    private static ArrayList<String> stringTokenizer(String options) throws ParseException{
+        
+        ArrayList<String> tokens = new ArrayList<String>();
+        final StringBuilder runningString = new StringBuilder(options.length());
```
cj81499, line 151: It seems like overkill to set the `capacity` of the `StringBuilder` to the length of `options`. In the case where there are many key-value pairs in  `options`, the length of each pair is a small fraction of the full length.

Is there something wrong with using the [no-arg constructor](https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#StringBuilder()) (which defaults capacity to 16) and let the `StringBuilder` raise the capacity as it needs?

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
```diff
@@ -129,4 +136,72 @@ private static Properties parseOptions(String ops) throws ParseException
...
+
+    /**
+     * Seperates key and value(possible quoted) pairs into sets
+     * Ex: key1=val1,key2="val2",key3=val3 -> [key1=val1, key2=val2, key3=val3]
+     * @param options option string containing all the options.
+     * @return ArrayList<String> seperated key=value pairs.
+     * @throws ParseException
+     */
+
+    private static ArrayList<String> stringTokenizer(String options) throws ParseException{
```
cj81499, line 148: How would this function handle input like: `key1="abc"123,key2=abc"123"`?

I'd think we'd want a `ParseException`, but it seems like it'd give `["key1=\"abc\"123", "key2=abc\"123\""]`.

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
```diff
@@ -33,14 +31,24 @@
...
-    
+
+    private enum ParseState
+    {
+        UNQUOTED,
+        QUOTED,
+        QUOTED_ESCAPED
+    }
+
+    private final static char QUOTE_CHARACTER = '\"';
```
cj81499, line 49: Should we also consider single quotes (`key1='val1'`)?

If so, we need to be careful not to exit on the wrong character when handling strings like `" 'hi' "` and `' "hello" '`

tli-ocient: We are only reserving `"` as our quote character. `'` is treated like any other.

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
tli-ocient: It would, but then it will fail to match with the ARG_VAL pattern in the following step and give a parse exception. Like so
Screen Shot 2021-12-10 at 1 06 42 PM

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
tli-ocient: I was worried about having to grow the capacity. Either way, I don't expect this to be a big deal speed or memory wise.  I changed it

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1352,10 +1354,14 @@ private static String scrubCommand(final String cmd)
...
 				}
-				if ((c == '\'' || c == '"') && (quote == '\0' || quote == c)) // char is an active quote
+				// Check for the start of a quote or the ending of a current quote.
+				if ((c == '\'' || c == '"') && (quote == '\0' || quote == c) && !inActiveEscape) // char is an active quote
 				{
 					quote = quote == '\0' ? c : '\0';
 				}
+				// If the current character is a backslash and not already in an active escape,
+				// then we are entering an active escape. Otherwise, we leave activeEscape.
+				inActiveEscape = c == '\\' && !inActiveEscape ? true : false;
```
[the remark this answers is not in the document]
tli-ocient: changed

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
cj81499: there's about a zero percent chance of me understanding how `ARG_VAL_PATTERN` works, so I'll take your word for it 😆

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
```diff
@@ -129,4 +136,72 @@ private static Properties parseOptions(String ops) throws ParseException
...
+                            runningString.setLength(0);
+                        }
+                    } else {
+                        // Otherwise, just add it to the current string
+                        runningString.append(currentChar);
+                    }
+                    break;
+                }
+                case QUOTED:{
+                    if(currentChar == ESCAPE_CHARACTER){
```
rveitch-Ocient, line 175: This implies escapes can only happen in quoted values, I think that's ok but want to confirm.

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
```diff
@@ -109,12 +117,11 @@ private static Properties parseOptions(String ops) throws ParseException
         }
         
         Map<String, String> keyValuePairs = new HashMap<String,String>();
-
-        String[] tokens = ops.split(TOKEN_SPLITTER_STRING, -1);
+        ArrayList<String> tokens = stringTokenizer(ops);
```
rveitch-Ocient, line 120: Java nit, you can swap most of the `ArrayList<String>` in this PR for `List<String>`

src/test/java/com/ocient/cli/extract/CliTest.java
```diff
@@ -0,0 +1,43 @@
...
+    private final String EXTRACT_RIGHT = ") AS select c1 + 1, c1 from sys.dummy10";
+    // Test which will ensure scrub command does not get broken for extract.
+    @Test
+    @Parameters({
+        "file_prefix=\"/ocient/db/clone_xgsrc/xgsrc/testExtract/results/result\"", // Simple path prefix option
+        "field_delimiter=\"\\\"\"", // Option with escaped quote
+        "field_delimiter=\" \"",
+        "field_delimiter=\"\"",
+        "record_delimiter=unquoted",
+        "field_delimiter=\"'with single quotes'\""
```
rveitch-Ocient, line 26: Do we have negative tests somewhere to flesh out cases like leading comma, double comma, etc?

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
tli-ocient: Yup. Furthermore, unquoted values should only be alphanumeric.

src/test/java/com/ocient/cli/extract/CliTest.java
tli-ocient: These tests are  to make sure the `scrubCommand` function in the CLI does not get broken. The tests you mentioned should go into `ExtractSyntaxParserTests.java`. Though they don't seem to be there so I will add them

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
tli-ocient: changed

src/main/java/com/ocient/cli/extract/ExtractSyntaxParser.java
```diff
@@ -129,4 +136,72 @@ private static Properties parseOptions(String ops) throws ParseException
...
+                    if(currentChar == QUOTE_CHARACTER){
+                        // We enter quoted mode
+                        runningString.append(currentChar);
+                        currentState = ParseState.QUOTED;
+                    } else if (currentChar == COMMA_CHARACTER){
+                        if(runningString.length() > 0){
+                            // We finished a string. Add it then clear the curentString
+                            tokens.add(runningString.toString());
+                            runningString.setLength(0);
+                        }
```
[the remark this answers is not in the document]
tli-ocient: Yea I think it makes sense to throw a ParseException here instead of just swallowing. Added

src/test/java/com/ocient/cli/extract/CliTest.java
tli-ocient: Scratch that. I must have screwed something up on this branch and `ExtractSyntaxParserTests.java` became untracked. I don't know how I managed to do that, but rather than performing git surgery I would like to merge without adding this test and then add it in a follow up PR. I have other small things that need to be added anyway.

dcoombs-ocient (approved):
LGTM. Thanks for the reviews @cj81499 @rveitch-Ocient

PR #219, opened 2021-12-20
Disable connection caching for connections 

tli-ocient:
See comment added in code. This isn't a performance concern since that particular flow requires human interaction.


src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2196,23 +2197,37 @@ public SQLWarning getWarnings() throws SQLException
 		return retval;
 	}
 
+	@Override
+	public boolean equals(Object o) {
```
jwilliams-ocient, line 2201: Add a comment saying `hashCode() & equals()` should almost ALWAYS be autogenerated, and to only roll your own if you:
1. have a legitimate reason to do so (v rare) 
2. know what you're doing
3.  write unit tests specifically for object equality

userdoc/release_notes.adoc
```diff
@@ -11,6 +11,7 @@ Below is the release notes for every driver version that has customer implicatio
 New Features::
 
  * Data extract tool released.
+ * Deactive connection caching for SSO flow.
```
jwilliams-ocient, line 14: Maybe specify the caching is disabled for calls to `Driver.createConnection(...)` (as opposed to `Connection.createStatement(...)`)

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1794,18 +1795,18 @@ public Struct createStruct(final String arg0, final Object[] arg1) throws SQLExc
...
 
-	@Override
-	public boolean equals(final Object o)
-	{
-		if (!(o instanceof XGConnection))
-		{
-			return false;
-		}
+	// @Override
+	// public boolean equals(final Object o)
```
jwilliams-ocient, line 1799: delete comments

src/main/java/com/ocient/jdbc/JDBCDriver.java
```diff
@@ -291,9 +291,11 @@ public Connection createConnection(final String hostname, final int portNum, fin
 						conn.connect();
 						LOGGER.log(Level.INFO, "Successfully connected");
 						connected = true;
-						synchronized (seenConnections)
-						{
-							seenConnections.put(conn, conn);
+						if(!conn.isDisabledConnCaching()){
```
jwilliams-ocient, line 294: this check is redundant?

jwilliams-ocient: hmm, maybe not. but something is fishy here...

jwilliams-ocient: yea, I think it is redundant.

1. conn_exist  ❌     |    caching  ❌      => `do_connect = true`
2. conn_exist  ✅    |    caching  ❌       => `do_connect = true`
3. conn_exist  ❌    |    caching  ✅       => `do_connect = true`
4. conn_exist   ✅   |    caching  ✅       => `do_connect = false`

jwilliams-ocient: Do (1) and (2) reduce to the same case? Maybe that's my confusion... Like (2) should _never_ happen

tli-ocient: Hmm, if no connection exists and caching is disabled (case 1), then do_connect is true and we add this to the cache(without that second check)

tli-ocient: then I think 2 will happen?

jwilliams-ocient: I guess that's where I'm confused. What other pieces of code update `seenConnections`?

tli-ocient: nowhere else, just here

jwilliams-ocient: and is there a problem if that happens w.r.t. the cache in `XGStatement`. 

RE `hashCode()` & `equals(...)` for `XGConnection`.

tli-ocient: I don't think so. I'm just concerned about us adding connections unecessarily to the `seenConnections` cache

jwilliams-ocient: why? Not sure I understand that. AFAICT we are strictly "reducing" the number of elements in `seenConnections`?

tli-ocient: Thats not true, we aren't removing any connections from here, ever. Only adding

jwilliams-ocient: Also, the statement pooling in `XGStatement` would _still_ broken in SSO contexts...

userdoc/release_notes.adoc
tli-ocient: changed comment

src/main/java/com/ocient/jdbc/XGConnection.java
tli-ocient: done

src/main/java/com/ocient/jdbc/JDBCDriver.java
tli-ocient: talked in zoom. this was necessary

PR #228, opened 2022-01-15
TOOLS-1018 update force external

ems236:
Force external didn't work in GQR tests because it didn't work across redirects.
Makes force external more stateful and act like the other set params

DB PR is here https://github.com/Xeograph/xgsrc/pull/20496


src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3117,6 +3107,12 @@ private void resendParameters()
 			} else {
 				setPriority(0.0, true);
 			}
+
+			if (forceExternal){
+				forceExternal(forceExternal);
+			} else{
+				//there shouldn't ever be a situation where resending params needs to turn this off again
```
ems236, line 3114: If you think it's better I can always send a false / reset here.

src/main/proto/ClientWireProtocol.proto
```diff
@@ -767,6 +768,9 @@ message SetParameter
 	message MaxTempDiskLimit{
 		int64 tempDiskLimit = 1;
 	}
+	message ResultSetColumnLimit{
```
ems236, line 771: Proto file was out of sync with db

tli-ocient (approved):
LGTM

PR #237, opened 2022-03-11
[DB-19576]: Added new columns in system queries

susmita-saha:
* sys.queries: initial_priority, initial_effective_priority, effective_priority, priority_adjust_factor, priority_adjust_time
* sys.completed_queries: initial_priority, initial_effective_priority, final_effective_priority, priority_adjust_factor, priority_adjust_time


src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1068,11 +1068,13 @@ else if (o != null)
...
-		System.out.format("%-40s%-15s%-15s%-20s%-20s%-15s%-20s%-15s%s%n", "query id", "user", "importance", "estimated time", "elapsed time", "status", "server", "database", "sql");
-		System.out.println(new String(new char[170]).replace("\0", "-"));
+		System.out.format("%-40s%-15s%-20s%-20s%-20s%-20s%-20s%-15s%-20s%-15s%s%n", "query id", "user", "eff_priority", 
+			"initial_priority", "initial_eff_priority", "estimated time", "elapsed time", "status", "server", "database", "sql");
+		System.out.println(new String(new char[210]).replace("\0", "-"));
 		for (final SysQueriesRow row : queries)
 		{
-			System.out.format("%-40s%-15s%-15s%-20s%-20s%-15s%-20s%-15s%s%n", row.getQueryId(), row.getUserid(), row.getImportance(), row.getEstimatedTimeSec(), row.getElapsedTimeSec(),
+			System.out.format("%-40s%-15s%-20s%-20s%-20s%-20s%-20s%-15s%-20s%-15s%s%n", row.getQueryId(), row.getUserid(), 
+				row.getEffectivePriority(), row.getInitialPriority(), row.getInitialEffectivePriority(), row.getEstimatedTimeSec(), row.getElapsedTimeSec(),
```
susmita-saha, line 1077: I don't this function being used anywhere.

src/main/proto/ClientWireProtocol.proto
```diff
@@ -408,7 +408,7 @@ message LocalQueriesResponse
 message SysQueriesRow
 {
 	string query_id = 1;
-	float importance = 2;
+	float effective_priority = 2;
```
tli-ocient, line 411: You cannot change the member name of protos. This will break backwards compatibility

src/main/proto/ClientWireProtocol.proto
```diff
@@ -451,7 +455,7 @@ message CompletedQueriesRow
 	uint64 timestamp_exec_start = 13;               // ms since epoch representation of the time the query began execution
 	string time_exec_start = 14;                    // the string representation of the time the query began execution	    
 	string uuid = 15;                               // the query plan uuid
-	double plan_priority = 16;                      // the priority assigned to this plan
+	double final_effective_priority = 16;           // the final effective dynamically adjusted priority
```
tli-ocient, line 458: Same here, cannot be changed.

src/main/java/com/ocient/cli/CLI.java
tli-ocient: Like you said, you shouldn't need to change this to have your new columns be reflected. Did you test it? The new columns should automatically be reflected once you change the protos.

tli-ocient: In fact, please delete this function.

tli-ocient (changes requested):

src/main/proto/ClientWireProtocol.proto
susmita-saha: I think it is okay to rename the fields, as long as I am not changing the type and position.
It shouldn't break backward compatibility.

src/main/java/com/ocient/cli/CLI.java
susmita-saha: Yes, I tested it, and the new columns show up in result set.

tli-ocient (approved):

PR #243, opened 2022-03-27
User/tli/col name pattern null map

tli-ocient:
https://ocient.atlassian.net/browse/DB-20088

Looks like we just missed this case in `fetchSystemMetadata`. Look at surrounding code.


SRKnoll (approved):
Well, from the surrounding, it seems like handling this for columns was missing. 
To me, it seems like '%' represents a NULL value or empty value for a column name. An empty columns name is odd, as it should not exist in a database. And for null values, a dedicated flag/boolean would be more appropriate.
Anyways, its the same for tables and schemas, so at least it is consistent.

PR #244, opened 2022-04-01
User/tli/set timestamp three args

tli-ocient:
https://ocient.atlassian.net/browse/DB-20111


src/main/java/com/ocient/jdbc/XGPreparedStatement.java
```diff
@@ -849,8 +849,20 @@ public void setTimestamp(final int parameterIndex, final Timestamp x) throws SQL
...
+			LOGGER.log(Level.WARNING, "setTimestamp() is throwing CALL_ON_CLOSED_OBJECT");
+			throw SQLStates.CALL_ON_CLOSED_OBJECT.clone();
+		}
+
+		while (parameterIndex - 1 > parms.size())
+		{
+			parms.add(null);
+		}
+		
+		parms.add(parameterIndex - 1, ((XGTimestamp) x).addMs(cal.getTimeZone().getOffset(x.getTime())));
```
jwilliams-ocient, line 864: this should be subtract (inbound v outbound)

PR #247, opened 2022-04-05
User/tli/turn off pooling option

tli-ocient:
https://ocient.atlassian.net/browse/DB-20110


jwilliams-ocient:
I'm pretty sure you said it didn't, but just to double check, updating this config file does not work, right? https://github.com/metabase/metabase/blob/master/resources/c3p0.properties

PR #257, opened 2022-05-27
DB-18053: Added OUTPUT SESSION command to CLI commands

Mryan918:
Added `OUTPUT SESSION <filename>` as a command in CLI.java. This starts a session in the designated file so the future commands, queries, query results, and timing info... etc will be stored in a file as you continue make queries and type in other commands. This session can be ended by typing `OUTPUT QUERY stdout`.

This command makes use of the System.setOut() function to redirect all print statements to both the standard window and the designated file using the created TeeStream class. This works in the same way as the tee command in linux.

Example use by running driver:

```
Ocient> OUTPUT SESSION /tmp/example_session.csv;
Ocient> SELECT * FROM sys.dummy1;
c1         
-----------
1 

Fetched 1 row
Ocient> OUTPUT SESSION stdout;
```
The file /tmp/example_session.csv will contain everything past the first line where OUTPUT SESSION was called, including the prompt, command, and result of the query typed in exactly as shown. This session can be continued for further commands, but in this case is ended when `OUTPUT SESSION stdout` is entered.

This is a good option compared to running `OUTPUT NEXT QUERY <filename>` multiple times in a row. Once you start the session, you can keep running queries and other commands and they will continue to be printed to both the file and the standard output console.


src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1763,4 +1801,68 @@ private static void update(final String cmd)
...
+			try {
+				PrintStream standard = System.out;
+				File logFile = new File(outputSessionFile);
+				PrintStream logOut = new PrintStream(new FileOutputStream(logFile, true));
+				System.setOut(logOut);
+				System.setErr(logOut);
+				System.out.println("Ocient> " + t);
+
+				/*
+				PrintStream teeStdOut = new TeeStream(System.out, logOut);
```
jwilliams-ocient, line 1815: nit: delete dead code

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1763,4 +1801,68 @@ private static void update(final String cmd)
...
+				PrintStream teeStdErr = new TeeStream(System.err, logOut);
+				*/
+
+				System.setOut(standard);
```
jwilliams-ocient, line 1819: `System.setOut` is called twice in this method?

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1092,6 +1122,7 @@ private static void printResultSet(final ResultSet rs, final ResultSetMetaData m
 			}
 
 			System.out.println(line);
+			//printSBToSession(line);
```
jwilliams-ocient, line 1125: nit: remove dead code

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1102,6 +1133,7 @@ private static void printResultSet(final ResultSet rs, final ResultSetMetaData m
...
```
jwilliams-ocient, line 1136: nit: remove dead code

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -981,6 +984,33 @@ private static void outputNextQuery(final String cmd)
...
+			
+			if(outputSessionFile.toLowerCase().equals("stdout")) {
+				System.setOut(new PrintStream(new FileOutputStream(FileDescriptor.out)));
+			} else {
+				final PrintStream standard = System.out;
+				PrintStream standarderr = System.err;
+				File logFile = new File(outputSessionFile);
+				PrintStream logOut = new PrintStream(new FileOutputStream(logFile, true));
+				PrintStream teeStdOut = new TeeStream(standard, logOut);
+				//PrintStream teeStdErr = new TeeStream(standarderr, logOut);
```
jwilliams-ocient, line 1000: nit: remove dead code

jwilliams-ocient: is removing `stderr` what fixed the duplicated messages?

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -981,6 +984,33 @@ private static void outputNextQuery(final String cmd)
 		}
 	}
 
+	private static void outputSession(final String cmd) {
+		try 
+		{
+			outputSessionFile = cmd.substring("OUTPUT SESSION ".length()).trim();
+			
+			if(outputSessionFile.toLowerCase().equals("stdout")) {
+				System.setOut(new PrintStream(new FileOutputStream(FileDescriptor.out)));
```
jwilliams-ocient, line 993: // When "stdout" is provided. The existing session log will be closed.
```suggestion
				System.setOut(new PrintStream(new FileOutputStream(FileDescriptor.out)));
```

jwilliams-ocient: Shouldn't the PrintStream object be closed?

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -981,6 +984,33 @@ private static void outputNextQuery(final String cmd)
...
+			} else {
+				final PrintStream standard = System.out;
+				PrintStream standarderr = System.err;
+				File logFile = new File(outputSessionFile);
+				PrintStream logOut = new PrintStream(new FileOutputStream(logFile, true));
+				PrintStream teeStdOut = new TeeStream(standard, logOut);
+				//PrintStream teeStdErr = new TeeStream(standarderr, logOut);
+				
+				System.setOut(teeStdOut);
```
jwilliams-ocient, line 1002: Same here - if an existing session log was open we need to close it to avoid a resource leak

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1763,4 +1801,68 @@ private static void update(final String cmd)
...
+				*/
+
+				System.setOut(standard);
+				System.setErr(standard);
+			} catch(Exception e) {
+				System.out.println("Error: " + e.getMessage());
+			}
+		}
+	}
+	private static void printSBToSession(StringBuilder t) {
```
jwilliams-ocient, line 1826: Is this used anywhere?

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1763,4 +1801,68 @@ private static void update(final String cmd)
...
+
+				System.setOut(standard);
+				System.setErr(standard);
+			} catch(Exception e) {
+				System.out.println("Error: " + e.getMessage());
+			}
+		}
+	}
+
+	public static class TeeStream extends PrintStream {
```
jwilliams-ocient, line 1844: I believe `close()` needs to be overriden so `this.out` can be closed

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1763,4 +1801,68 @@ private static void update(final String cmd)
...
+		public TeeStream(PrintStream out1, PrintStream out2) {
+			super(out1);
+			this.out1 = out1;
+			this.out = out2;
+		}
+		@Override
+		public void write(byte buf[], int off, int len) {
+			try {
+				out1.write(buf, off, len);
+				out.write(buf, off, len);
```
jwilliams-ocient, line 1857: So invoking the super class did not work, what happens when one of the other `PrintStream#write(...)` method signatures get called? Does this method get called?

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1763,4 +1801,68 @@ private static void update(final String cmd)
...
+			} catch (Exception e) {
+				System.out.println("Error: " + e.getMessage());
+			}
+		}
+		@Override
+		public void flush() {
+			super.flush();
```
jwilliams-ocient, line 1864: Here we are invoking the superclass but not in `write(...)`?

jwilliams-ocient (commented):
Nice job getting this up! Looks good overall, mostly "nits" (minor issues) around dead code, but a couple of other questions

daveboutcher:
One day I will add system tests for the CLIs so that JDBC, python, and ODBC are consistent.  Today is not that day though :-)

src/main/java/com/ocient/cli/CLI.java
Mryan918: No, it was removed in my latest version that was cleaned up. Sorry about that

src/main/java/com/ocient/cli/CLI.java
Mryan918: No, it was regarding the write method in the TeeStream class. It was originally calling super() instead of out1.write and it was recognizing super and out2 to be the same thing, so it was printing twice to the file each time.

src/main/java/com/ocient/cli/CLI.java
Mryan918: Yes, I call it the first time to setOut to the session and then print it there, and then I set it back to the standard output window. This was a work around because the prompt wasn't technically an PrintStream so it wasn't printing that to the session

daveboutcher:
One thought (@jwilliams-ocient, we should talk about this.)  For pyocient we have been trying to make all the commands (CLI and otherwise) consistent with SQL syntax.  for the `source` statement, the path (in pyocient) needs to be quoted.  So 
`source '/tmp/myfile.sql';` because that is more consistent. I am aware we have never tried to be syntactically consistent in the JDBC driver. Since this is a new command, would we consider requiring the path to be quoted?

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -981,6 +984,33 @@ private static void outputNextQuery(final String cmd)
 		}
 	}
 
+	private static void outputSession(final String cmd) {
+		try 
+		{
+			outputSessionFile = cmd.substring("OUTPUT SESSION ".length()).trim();
```
jwilliams-ocient, line 990: > we have been trying to make all the commands (CLI and otherwise) consistent with SQL syntax. for the source statement, the path (in pyocient) needs to be quoted. So source '/tmp/myfile.sql'; because that is more consistent...

Dave brings up a good point here, I think it's reasonable to enforce consistency moving forward. The user should (single) quote the filename here.

jwilliams-ocient: Do we need to trim trailing whitespace as well?

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -981,6 +985,34 @@ private static void outputNextQuery(final String cmd)
...
+			if((outputSessionFile.length() > 2)  && (outputSessionFile.startsWith("'")) && (outputSessionFile.endsWith("'"))) {
+				outputSessionFile = outputSessionFile.substring(1, outputSessionFile.length() - 1);
+
+				if(outputSessionFile.toLowerCase().equals("stdout")) {
+					System.setOut(new PrintStream(new FileOutputStream(FileDescriptor.out)));
+					fileOut.close();
```
jwilliams-ocient, line 997: Great that we are now closing this resource! 

This ensures the backing file descriptor is closed, but leaves the `PrintStream` object "alive". While `fileOut` is still technically a valid java object, there really isn't a purpose to keep the reference around.

`fileOut`, while technically a valid java object isn't valid in any real sense. Any operation would throw some form kind "This stream is closed" exception. By nullifying `fileOut` we maintain an implicit guarantee that `fileOut` is **_always_** an open handle (comment would future us!). This is a defensive programming technique.

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1763,4 +1801,46 @@ private static void update(final String cmd)
...
+		public TeeStream(PrintStream out1, PrintStream out2) {
+			super(out1);
+			this.standardOutput = out1;
+			this.fileOutput = out2;
+		}
+		@Override
+		public void write(byte buf[], int off, int len) {
+			try {
+				standardOutput.write(buf, off, len);
+				fileOutput.write(buf, off, len);
```
jwilliams-ocient, line 1835: Assuming `fileOutput` is the passed to the super constructor, I still don't understand why this doesn't work
```suggestion
				super.write(buf, off, len);
```

Mryan918: Just changed it, this does work! I'm not sure why this doesn't work with standardOutput (which is what I originally used)

PR #261, opened 2022-06-01
User/jwilliams/extract skip header

jwilliams-ocient:
# Overview 

## Ticket
https://ocient.atlassian.net/browse/DB-20755

## Background
Dearborn has the need to extract data to files with the JDBC driver. This has the option today to:
```
skip_header=true
```
that will avoid setting a header.

When left as `skip_header=false`, a header is rendered to the file. This header is included on every file that is generated in the split.

# Changes
Two updates are required for this feature.

New setting to only include the header on the FIRST file in the set of files created by the extract:
```
header_mode=first_file
```
In this mode, the customer would prefer that the number of records included on the first file is reduced by one so that all files include the same number of rows (with the exception of the final file which will usually have fewer rows).

For example: 
```
for max_rows_per_file = 100k and 350k rows total

File 1: 1 header row, 99,999 data rows (100k rows total)

File 2: No header, 100k data rows

File 3: No header, 100k data rows

File 4: No header, 50,001 data rows
```
By default, when `header_mode=all_files`, the extract can continue operating as it does today. 

For example:
```
All files in split output: 1 header row, 100k data rows (last file will have fewer rows when data is done)
```


src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
```diff
@@ -160,8 +178,8 @@ public ExtractConfiguration(final Properties properties)
     public static final String FIELD_DELIMITER = "field_delimiter";
     public static final String DEFAULT_FIELD_DELIMITER = ",";
 
-    public static final String SKIP_HEADER = "skip_header";
-    public static final boolean DEFAULT_SKIP_HEADER = false;
+    public static final String HEADER_MODE = "header_mode";
+    public static final HeaderMode DEFAULT_HEADER_MODE = HeaderMode.NONE;
```
zzpxyx, line 182: The old code has `false` as the default value. Do we want the new code to have the same default, i.e. HeaderMode.ALL_FILES?

zzpxyx (commented):
LGTM overall. One comment/question below.

src/main/java/com/ocient/cli/extract/ExtractConfiguration.java
jwilliams-ocient: Good question, JK brought this up too. He and Selva agreed that "none" should be the default value here

PR #270, opened 2022-06-27
DB-19233: Adding Pretty Print Table and Json Formatting

Mryan918:
Added Pretty Print formatting for Table and Json Formatting in PrintResultSet in src/main/java/com/ocient/cli/CLI.java


pom.xml
```diff
@@ -86,6 +86,11 @@
       <artifactId>commons-io</artifactId>
       <version>2.11.0</version>
     </dependency>      
+    <dependency>
```
jwilliams-ocient, line 89: ended up not needing gson?

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1067,92 +1075,60 @@ else if (o != null)
...
-				else
-				{
-					line.append("\n");
-				}
-			}
-
-			if (line.length() != 0)
-			{
-				System.out.println(line);
+				List<String[]> data = new ArrayList<>();
```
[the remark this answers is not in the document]
jwilliams-ocient: formatting is still off here...

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -1128,92 +1136,60 @@ else if (o != null)
...
+				JsonArray jsonArray = new JsonArray();
+        		while (rs.next()) {
+         		   	int total_columns = meta.getColumnCount();
+        	    	JsonObject obj = new JsonObject();
+          		  	for (int i = 0; i < total_columns; i++) {
+						Object curr = rs.getObject(i+1);
+						if(curr == null) {
+							curr = "NULL";
+						}
+						else if(curr instanceof XGArray) {
```
jwilliams-ocient, line 1178: I think we need to test with the `all_data_types` schema to make sure there aren't any other types we need to add

jwilliams-ocient:
```
# kill running rolehostds
pkill -9 rolehostd

# create a cluster and load the `all_data_types` schema
bazel run runner -- -m scripts/systemtesting/test_systemCorrectness.py --schema all_data_types -S
```

jwilliams-ocient: I'm fairly certain you'll have to do something similar for All of the `XG*` classes. See `getObject(int)` implementation

pom.xml
Mryan918: Gson dependencies are on lines 64-68 of the pom file

PR #271, opened 2022-06-27
fix: escape escape character for non-comma delimited extracts

jwilliams-ocient:


src/test/java/com/ocient/cli/extract/writer/CustomTsvWriterTest.java
```diff
@@ -151,4 +151,26 @@ public void testEnclosureRequired() {
         assertEquals("\"a\"|\"b\"|\"c|d\"|\"ef\"\"crazy\"\"\"\n",
                 byteArrayOutputStream.toString(StandardCharsets.UTF_8));
     }
+
+    /**
+     * Escape: \\\\
```
zzpxyx, line 156: I think this may be a bit confusing. I'd recommend putting the _real_ char here:

```suggestion
     * Escape: \
```

src/test/java/com/ocient/cli/extract/writer/CustomTsvWriterTest.java
```diff
@@ -151,4 +151,26 @@ public void testEnclosureRequired() {
...
+     * Delimiter: |
+     * Enclosure: "
+     * Force enclosure: true
```
zzpxyx, line 159: I think by default this setting is false.

```suggestion
```

PR #289, opened 2022-09-20
DB-22035: Add START PIPELINE and STOP PIPELINE

pyusimocient:
https://ocient.atlassian.net/browse/DB-22035


jwilliams-ocient (approved):
Approved under the assumption that these commands are a part of the Ocient SQL Engine dialect

PR #290, opened 2022-09-21
DB-22039: ensure backwards compatibility with JRE8

jwilliams-ocient:
# Overview
So this is pretty annoying... we need to ensure JRE8 compatibility but cannot enable maven compiler checks because of this bit.
https://github.com/Xeograph/ocient_jdbc/blob/6f79f0413c83af879d400e2760d33283fc6d8ba6/src/main/java/com/ocient/jdbc/XGConnection.java#L2198-L2202

# Test
In order to test JRE8 compatibility, add the following properties to the pom file.
```
<properties>
  ...
  <maven.compiler.source>1.8</maven.compiler.source>
  <maven.compiler.target>1.8</maven.compiler.target>
  ...
</properties>
```


benschwartz10:
Looks good to me! I tested the branch and encountered no errors

PR #292, opened 2022-09-23
Make JDBC connection properties case-insensitive

benschwartz10:
Added a new class CaselessProperties which extends the java.util.Properties class, making the keys case-insensitive.

Converted the rest of the connection properties to the ConnectionProperty enum


src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -406,7 +423,7 @@ public OAuthFlow getOAuthFlow(Properties properties, OAuthFlow defaultValue) thr
 		}
 
 		public HandshakeType getHandshakeType(Properties properties, HandshakeType defaultValue) throws SQLException {
```
jwilliams-ocient, line 425: This was a poor design choice on my part... You shouldn't be forced to provide a default value here, or in any of the other getter methods. This might be a better approach for the get-or-default pattern.
```suggestion
		public <E> Optional<HandshakeType> getHandshakeType(Properties properties) throws SQLException {
```
or
```
                 public OptionalInt getInt(final Properties properties) {
```
That way, the caller can do
```
final int maxRows = ConnectionProperty.MAX_ROWS.getInt(properties).orElse(0);
if((proposedMaxRows < 1) && (proposedMaxRows != -1)){
  // throw
}
```
Thoughts?

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3702,64 +3719,64 @@ public void releaseSavepoint(final Savepoint arg0) throws SQLException
...
 			if(proposedNetworkTimeout <= 0){
-				throw SQLStates.INVALID_ARGUMENT.cloneAndSpecify(String.format("network timeout must be greater than 0, specified: %d", proposedNetworkTimeout));
+				throw SQLStates.INVALID_ARGUMENT.cloneAndSpecify(String.format("%s must be greater than 0, specified: %d", ConnectionProperty.NETWORK_TIMEOUT.key(), proposedNetworkTimeout));
 			}
 		}
-		if (properties.containsKey("longQueryThreshold") && properties.get("longQueryThreshold") != null)
+		if (properties.getProperty(ConnectionProperty.LONG_QUERY_THRESHOLD.key()) != null)
 		{
-			int proposedSetPso = Integer.parseInt((String) properties.get("longQueryThreshold"));
+			int proposedSetPso = ConnectionProperty.LONG_QUERY_THRESHOLD.getInt(properties, -2);
```
jwilliams-ocient, line 3761: Is `-2` intentional here?

jwilliams-ocient: totally weird that you have to specify a default value here, apologies for that...

benschwartz10:
Original commit was not verified. Used a force push to fix this

src/main/java/com/ocient/util/CaselessProperties.java
```diff
@@ -0,0 +1,46 @@
+package com.ocient.util;
+
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Reader;
+import java.util.Properties;
+import java.util.Set;
+
+public class CaselessProperties extends Properties {
```
jwilliams-ocient, line 9: Do we need to override the methods that return iterators, that bypass the `getProperty()` code path, like `keys()`, `values()`, `entrySet()`, etc.?

jwilliams-ocient: throwing `UnsupportedOperationException` is probably fine too

jwilliams-ocient: unless they are used in the copy constructor

jwilliams-ocient (commented):
Looks great, minor question about iterators

src/main/java/com/ocient/jdbc/XGConnection.java
benschwartz10: Yes! That was to trigger the error condition since that default value should never be used.

src/main/java/com/ocient/util/CaselessProperties.java
benschwartz10: Really these are things that should be handled by java.util.properties since they suffer the same problem. The only way I saw them dealing with it was to say in the [docs not to use those methods](https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html): 
> Because Properties inherits from Hashtable, the put and putAll methods can be applied to a Properties object. Their use is strongly discouraged as they allow the caller to insert entries whose keys or values are not Strings. The setProperty method should be used instead. If the store or save method is called on a "compromised" Properties object that contains a non-String key or value, the call will fail. Similarly, the call to the propertyNames or list method will fail if it is called on a "compromised" Properties object that contains a non-String key

This is something that could be fixed up in CaselessProperties though. My thought process now is to set them all to UnsupportedOperation because they shouldn't be used anyways.

jwilliams-ocient: agreed, we should override with `UnsupportedOperation`

jwilliams-ocient (approved):
Thanks for cleaning these properties up!

jwilliams-ocient:
there's actually one more step

jwilliams-ocient:
we need to:
1. bump the version https://github.com/Xeograph/ocient_jdbc/blob/a823086c4723ed6f6ceda3ea749533b117b2e34a/pom.xml#L6
2. add a release note here https://github.com/Xeograph/ocient_jdbc/blob/master/userdoc/release_notes.adoc

jwilliams-ocient (approved):

PR #293, opened 2022-09-28
[DB-21708] Add support for EXPORT PIPELINE in JDBC

jfeehery-ocient:


daveboutcher:
Don't forget ODBC and pyocient

jfeehery-ocient:
@daveboutcher Seems like ODBC and pyocient support commands that start with "EXPORT", whereas JDBC only supports "EXPORT TABLE" or "EXPORT VIEW", would it be better to switch JDBC to support `|| startsWithIgnoreCase(cmd, "EXPORT")` instead?

PR #297, opened 2022-10-18
Fix: remove error on connection property override

benschwartz10:
Removed an exception that would throw if one of the connection properties was already set and was attempted to be overwritten. Now, all connection properties can be overwritten.


jwilliams-ocient (commented):
LGTM, just needs a version bump + release note 👍🏾

benschwartz10:
Oops, forgot this again. I'll add it back

daveboutcher:
Can you wait to release it? I have another fix coming

jwilliams-ocient (approved):

PR #299, opened 2022-10-21
DB-22157: Set Service class at session level

pyusimocient:
https://ocient.atlassian.net/browse/DB-22157


daveboutcher (changes requested):
Wait!!!! We need to talk about this.  Part of the idea of service classes is that an administrator can set the service class for a user to restrict things like query time or priority.  if I read this correctly, this allows any user to just set the service class for a query to anything they want!!?!?!

daveboutcher (approved):

pyusimocient:
@daveboutcher's objections were withdrawn in an offline conversation and that is why this PR is approved by him.

PR #313, opened 2022-12-27
feat: add report logger to JDBC driver

jwilliams-ocient:
# Overview
Provide the ability to write JDBC metrics to a file.

## Changes
Two additional CLI commands have been added:
- `REPORT LOG ON <int>?`: enables metrics collection and redirects reporting events to the log4j-specified appender(s). The `<int>` argument is the reporting interval in seconds. The interval defaults to 30 seconds If no value is provided. 
- `REPORT LOG OFF`: disables the report logger.

## Log4j
The report log location is determined by log4j configuration, which can be injected into the application in many ways. The logger name is `com.ocient.util.ReportLogger`.

A default configuration is provided in the JAR we distribute, the default config can be retrieved by an archive utility like unzip. For example:

```shell
unzip -p <path-to-jar> log4j2.xml
```


alexclavel-ocient (approved):

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -2071,21 +2101,128 @@ public String next() {
...
+			// Default loggig interval is 30 seconds
+			int reportLogInterval = 30;
+
+			// Check for a specified reporting interval
+			final String reportLogIntervalStr = cmd.substring("REPORT LOG ON".length()).trim();
+			if (!reportLogIntervalStr.isEmpty()) {
+				try
+				{
+					reportLogInterval = Integer.parseInt(reportLogIntervalStr);
+					if (reportLogInterval < 0) {
```
alexclavel-ocient, line 2137: Do we want to allow interval of 0?

alexclavel-ocient (approved):

src/main/java/com/ocient/cli/CLI.java
jwilliams-ocient: no we definitely don't, good catch. The min interval should be 5 seconds

PR #318, opened 2023-02-24
DB-23045: Update gitactions checkout command to fetch all git history

awang-ocient:
### Before

JDBC Deploy workflow fails because `git-commit-id-plugin` expects full git history, but `actions/checkout@v2` in `jdbc-deploy.yml` only fetches the latest commit by default.

I reproduced the error locally by cloning this repo with  `--depth 1`. As expected, the maven build fails.

### After

The maven build works when I clone the repo normally with the full git history. I modified `jdbc-deploy.yml` so that the github workflow should fetch the entire git history.


jwilliams-ocient (approved):

jwilliams-ocient:
🔥 🔥

PR #319, opened 2023-03-01
DB-23453: Override "METRICS ON" command to allow logging metrics to user-specified directory

awang-ocient:
Provide the ability for user to specify a directory to write JDBC metrics to.

### Existing Functionality

`METRICS ON` : stream metrics to default port 8779
`METRICS ON <int>` : stream metrics to user-specified port

### New Functionality

`METRICS ON <alphanumeric>` : periodically log metrics to user-specified directory

The log filename is generated from the `query UUID` if non-null, otherwise a random alphanumeric string of length 10 is used.

The metrics are logged under `<alphanumeric_directory>/<filename>.log`

### Behavior

`METRICS OFF` will turn off both streaming metrics to a port and logging metrics to a file.

Users cannot stream metrics to a port and log metrics to a file at the same time. They must run `METRICS OFF` first before running a new `METRICS ON <int>` or  `METRICS ON <alphanumeric>` command.

### Miscellaneous

The default logging time interval is 30 seconds. The file rollover size is 100MB.


src/main/java/com/ocient/util/ReportLogger.java
```diff
@@ -77,6 +78,34 @@ public boolean isRunning() {
...
+
+    public String getMetricsFilePattern(String directory, String filename) {
+        return directory + "/" + filename +  "/%d{yyyy-MM-dd}-%i.log.gz";
+    }
+
+    public String generateRandomString() {
+        RandomStringGenerator generator = new RandomStringGenerator.Builder()
+													.withinRange('0', 'z')
+													.filteredBy(Character::isLetter, Character::isDigit)
+													.build();
```
alexclavel-ocient, line 104: Why not use `Character::isLetterOrDigit`([src](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isLetterOrDigit-int-))?

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -2125,36 +2144,32 @@ private static void reportLog(final String cmd, final boolean on) {
...
-					}
-				}
-				catch (final NumberFormatException e)
-				{
-					System.out.println("Invalid reportLogInterval specified for 'REPORT LOG ON'. Must be an integer value, greater than 5. " + e.getMessage());
-					lastCommandErrored = true;
-					return;
-				}
-			}
-
```
alexclavel-ocient, line 2150: Why do we not allow users to specify logging intervals anymore?

src/main/java/com/ocient/util/ReportLogger.java
```diff
@@ -77,6 +78,34 @@ public boolean isRunning() {
...
+        LOGGER.addAppender(rfa);
+    }
+
+    public String getMetricsFilePath(String directory, String filename) {
+        return String.format("%s/%s.log", directory, filename);
+    }
+
+    public String getMetricsFilePattern(String directory, String filename) {
+        return directory + "/" + filename +  "/%d{yyyy-MM-dd}-%i.log.gz";
+    }
```
alexclavel-ocient, line 98: These methods likely should have some null checking, and should make sure that the length of the string is not 0.

src/main/java/com/ocient/util/ReportLogger.java
awang-ocient: good catch

src/main/java/com/ocient/util/ReportLogger.java
```diff
@@ -77,6 +78,34 @@ public boolean isRunning() {
...
+
+    public String generateRandomString() {
+        RandomStringGenerator generator = new RandomStringGenerator.Builder()
```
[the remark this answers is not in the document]
awang-ocient: I can move this into the class constructor

src/main/java/com/ocient/cli/CLI.java
awang-ocient: We're replacing the `REPORT LOG` command with `METRICS` command and after speaking w Jordan + Dashang we decided having a default interval is sufficient/better user experience.

src/main/java/com/ocient/cli/CLI.java
```diff
@@ -2125,36 +2144,32 @@ private static void reportLog(final String cmd, final boolean on) {
...
 				REPORT_LOG_SERVICE = new ReportLogService();
 			}
 
 			metricsState.enable(ReportType.LOG);
 
+			// Remove file appenders configured in log4j2.xml
+			REPORT_LOG_SERVICE.reportLogger.removeAllFileAppenders();
+
+			final String directory = cmd.substring("METRICS ON".length()).trim();
+			final String filename =  ((XGStatement)stmt).getQueryId().orElse(REPORT_LOG_SERVICE.reportLogger.generateRandomString());
```
jwilliams-ocient, line 2160: While in theory, this _is_ a great idea and follows Dashang's request to a T, in practice this actually doesn't work in practice. It would be great to write a single log file per query but that doesn't really make sense here. 

This is a _session_ level logger, that is, metrics for _all_ queries executed in the CLI session will be routed to this appender. With that in mind, this should contain a _session_ level value, something like `XGConnection.sessionID`. The one thing I don't like about the session id is that you have to look at the CLI stdout text to find the most recent log file because the session id is a randomly generated UUID. Perhaps we should include _some_ temporal component in the name so that end users can find the most recent log pretty easily.

I think we should use either:
1. `<str>.log`: User-injected alphanumeric string as the filename
2. `<session_id>.log`: The CLI session id as the filename

jwilliams-ocient (commented):
Code looks great! one comment

src/main/java/com/ocient/cli/CLI.java
awang-ocient: Ok I think I'll go with sessionID. In my opinion it makes a better user experience since they don't have to input both a directory and filename.

jwilliams-ocient (approved):

PR #322, opened 2023-03-10
perf: refactor connection pooling architecture [DB-23575]

jwilliams-ocient:
# Overview
This PR introduces a connection pooling mechanism using the Apache Commons Object Pool library to reduce the number of times the driver needs to open a new socket connection to the database. 

## Background
There's a hand-rolled caching mechanism that exists in the driver today, but it isn't utilized for all db connections. As more requirements/features have been added to the driver, it has become increasingly difficult to reason about caching.

Prior to this PR, two unbounded caches exist:
- one for `XGConnection` objects, this lived in `JDBCDriver`
- one for `XGStatement` objects, this lived in `XGStatement`

My guess is the second cache was added because there was another use case where caching a connection would come in handy. Well, I have a third use case for pooling secondary Resultset connections. Architecturally speaking, I didn't see a reason to have disjoint db connection caches. Trying to manage one cache is hard enough! Cleaning this up is the primary motivation for this refactor.

The guiding principle is:
> Cache the most expensive resource(s) to create and JIT-provision everything else.

### XGConnection
In this case, connection objects should be cached because they are expensive to create. Creation involves:
1. TCP/SSL handshakes
2. Ocient Wire Protocol handshake

By keeping these connections cached, we can avoid a bit of redundant network traffic and computational work.

### XGStatement

The `XGStatement` object is just a wrapper around `XGConnection` that provides helper functions for communicating via Ocient Wire Protocol. There's no need to cache statement objects.

### Global Connection Pool
Now there is a single `XGConnection` pool per JVM instance.  All consumers including:
- statement creation
- background tasks like fetching the server version
- secondary resultset threads
 
"lease" objects from the connection pool and "return" them once `close()` is called on the connection.
```mermaid
graph TD;
  ConnectionPool -->|lease| Statement;
  ConnectionPool -->|lease| BG_Task;
  ConnectionPool -->|lease| RS_Thread;
  Statement -->|return| ConnectionPool;
  BG_Task -->|return| ConnectionPool;
  RS_Thread -->|return| ConnectionPool;
```
Importantly, all `XGConnection` objects in the cache have an active connection to the database. If a disconnect occurs, the connection will be evicted.

### Global Connection Pool Cache Key

Connections are grouped by the minimal set of defining connection properties. These are outlined in `XGConnectionInfo#equals()`. This deviates from the previous implementation that took into account session-level properties. These properties are restored when the connection is activated so there's no need to consider them is the equality check. https://github.com/Xeograph/ocient_jdbc/blob/1dfce3981d595bac7b0f0ddb78c2558dba25976d/src/main/java/com/ocient/jdbc/XGConnectionInfo.java#L414-L421

### Redirects
Redirects are tricky. The problem, in this case, is that two of the immutable connection attributes: `host` and `port` are changing. The Apache Commons Pool does not allow us to add a pre-allocated pooled object to it so once the redirect occurs, the associated `XGConnection` object will not be returned to the connection pool when closed. Notably, forked connections _are_ eligible for caching.
https://github.com/Xeograph/ocient_jdbc/blob/a405a86cac69420cd2c7ed0b194e77ef59bc37a8/src/main/java/com/ocient/jdbc/XGConnection.java#L3772-L3821

### Global Connection Pool Configuration
Some pool config options can be overridden via an environment variable. Please review these settings carefully. https://github.com/Xeograph/ocient_jdbc/blob/2e2069bc07f17b423f179c69e04341a48ac2ce8a/src/main/java/com/ocient/jdbc/XGConnection.java#L885-L931

### Global Server Version Cache
We now keep a global cache of server versions. The cache is invalidated on reconnect to handle the case where the server's `rolehostd` process is restarted with a new version.
https://github.com/Xeograph/ocient_jdbc/blob/a405a86cac69420cd2c7ed0b194e77ef59bc37a8/src/main/java/com/ocient/jdbc/XGConnection.java#L627-L655

### Detach From Query Wire Opcode
This opcode is used by secondary Resultset threads to drop its reference to the plan. 
https://github.com/Xeograph/ocient_jdbc/blob/a405a86cac69420cd2c7ed0b194e77ef59bc37a8/src/main/proto/ClientWireProtocol.proto#L248-L251
> **Warning**
> This might not be necessary at all... it looks like `m_plan->reset()` is called when the client sends the `CLOSE_RS` opcode, that might be sufficient. Defer to the cmdcomp guys on this. I'd prefer to remove this entirely if possible.

See https://github.com/Xeograph/xgsrc/pull/26636


awang-ocient:
We are using an apache pooling library so that the JDBC driver can utilize existing XGConnections from that "pool" before having to create a new one?

awang-ocient:
Does each XGConnection still only handle one query to the DB?

awang-ocient:
When leasing an XGConnection from the "pool", will any available XGConnection work, or does it depend on the use case? Curious about when new XGConnections are created vs. when we can use one from the pool.

awang-ocient:
Is this PR mainly for refactoring or will we see perf improvements as well?

jasonar81 (approved):

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -815,63 +887,218 @@ public void refreshToken() throws SQLException{
...
+	 *         the pool
+	 * @throws Exception if there is an error retrieving a connection from the pool
+	 */
+	public static XGConnection leaseConnection(XGConnectionInfo info) throws SQLException {
+		// Connections are leased here, and returned to the pool on close(). If the
+		// connection dies, or is otherwise, redurned invalid, the pooled connection
+		// is invalidated and immediately destroyed.
+		try {
+			return POOL.borrowObject(info);
+		} catch (Exception e) {
```
jwilliams-ocient, line 995: > @awang-ocient We are using an apache pooling library so that the JDBC driver can utilize existing XGConnections from that "pool" before having to create a new one?

Correct, there are a number of instances where the driver will need to communicate with the database via sideband channel. By "sideband channel", I mean a channel for sending requests not directly related to sending SQL queries and receiving Resultset data back from the server. 

A more robust protocol might use a duplex model so that the client and server can have multiple "conversations" happening at once over the same TCP socket. However, version 1 of the Ocient Wire Protocol is _simplex_, only one actor can send or receive requests at any given time. And even then, the protocol only allows the client to send requests. Unsolicited requests from the server are NOT permitted (an example of an unsolicited request might be the server telling the client about a new server that just came online, or propagating a WARNING from the server to the client outside of the context of query execution, etc.

But yes, to answer your question, yes, the goal here is to reuse existing db connections from the pool for:
- primary db channels
- secondary db channels (channels that are "attached" to an executing query)
- sideband channels

jwilliams-ocient: > Does each XGConnection still only handle one query to the DB?

correct, each query the user wants to submit to the database is scoped to a "statement", either a `java.sql.Statement` or a `java.sql.PreparedStatement` - the distinction between the two isn't too important in this context. 

A statement is:
- backed by exactly 1 TCP connection to the database
- has at most 1 active Resultset (query)

A simple case is the REPL in the CLI. When a `connect to...` command is entered, the process creates a connection to the db and creates a "statement". https://github.com/Xeograph/ocient_jdbc/blob/cc28d2c4513623edd75ac69d801984a74188776f/src/main/java/com/ocient/cli/CLI.java#L236-L240

A reference to the stmt object is captured in the REPL and is used to submit queries to the database sequentially, that is, the user can only submit 1 query at a time.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -815,63 +887,218 @@ public void refreshToken() throws SQLException{
...
+		// Lease a connection from the pool
+		return leaseConnection(info);
+	}
+
+	/**
+	 * FOR THE LOVE OF GOD, PLEASE DO NOT INCREASE VISIBILITY OF THIS CTOR
+	 * AND DO NOT MAKE ANOTHER CTOR. MAKE A STATIC CTOR THAT CLEARLY
+	 * DOCUMENTS WHATEVER YOU ARE TRYING TO TO IF YOU ABSOLUTELY HAVE TO.
+	 */
+	private XGConnection(final XGConnectionInfo info)
```
jwilliams-ocient, line 1065: > @awang-ocient Does each XGConnection still only handle one query to the DB?

correct, each query the user wants to submit to the database is scoped to a "statement", either a java.sql.Statement or a java.sql.PreparedStatement - the distinction between the two isn't too important in this context.

A statement is:
- backed by exactly 1 TCP connection to the database
- has at most 1 active Resultset (query)

A simple case is the REPL in the CLI. When a connect to... command is entered, the process creates a connection to the db and creates a "statement". https://github.com/Xeograph/ocient_jdbc/blob/cc28d2c4513623edd75ac69d801984a74188776f/src/main/java/com/ocient/cli/CLI.java#L236-L240

A reference to the stmt object is captured in the REPL and is used to submit queries to the database sequentially, that is, the user can only submit 1 query at a time.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4761,18 +5068,233 @@ public Session.State sendRefreshToken(final Session.State oldState) throws SQLEx
...
+					&& (conn.serverVersion == null || conn.serverVersion.isEmpty())) {
+				try {
+					conn.fetchServerVersion();
+				} catch (final Exception e) {
+				}
+			}
+
+			// set the "closed" flag to false, all systems go!
+			conn.closed = false;
+		}
```
jwilliams-ocient, line 5209: > @awang-ocient When leasing an XGConnection from the "pool", will any available XGConnection work, or does it depend on the use case? Curious about when new XGConnections are created vs. when we can use one from the pool.

Great question! The db connection _is_ stateful. There are both session-level and connection-level parameters that have to be restored when a connection is leased from the pool - similar to any context-switching mechanism. For this reason, we ensure that any leased connection was created by the same user, same protocol version, etc. (anything that is _not_ a "hint" in `XGConnectionInfo`).

when leased, the `activateObject()` method in the factory will restore any state the connection needs to move forward, including restoring:
- connection properties on both the client and server
- timeout configuration
- default schema
- any Single Sign-On session-related info

pom.xml
```diff
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.ocient</groupId>
   <artifactId>ocient-jdbc4</artifactId>
-  <version>2.100</version>
+  <version>2.101</version>
```
jwilliams-ocient, line 6: > @awang-ocient Is this PR mainly for refactoring or will we see perf improvements as well?

We expect to see performance enhancements when the application:
- repeatedly creates and tears down connections
- submits queries that return _unsorted_ Resultsets

The Continuous Query Runner (CQR) is an application that we use internally for concurrency testing. Its aim is to mimic the pattern of a customer application, for the BF workload, we are trying to increases queries-per-second (QPS). The CQR basically does this:
```
# Create a connection
conn = newConnection()

# Create 50 threads
for i in range(50):
  
  # Until the test is stopped...
  while True:

    # Create a "new" statement
    with conn.createStatement() as stmt:
  
      # Execute a query that returns an unsorted RS
      results = stmt.executeQuery(...)

      # No-op process results
      process(results)
```
Prior to this PR, the "Create a 'new' statement" part returned a cached statement (no change here) but the "Execute a query that returns an unsorted RS" _always_ spawned up to 32 _new_ connections, then immediately tore then down.

After this PR, any additional connection needed for "Execute a query that returns an unsorted RS" will be pooled.

jwilliams-ocient: There's a secondary issue of ensuring the client does not chew through the operating system's file descriptor limit. Each TCP socket requires an fd. I am currently looking at using some version of a Heirarchical Token Bucket (HTB) algorithm to ensure that:
- we don't exceed the `ulimit`
- we provide fairness amongst open client-server links when client is "busy"
- we maximize utilization of bandwidth in idle times

PR #324, opened 2023-03-20
perf: pool resultset threads [DB-23575, DB-23660]

jwilliams-ocient:
## Overview
The existing implementation spawns N threads when creating a resultset where N is the number of fetch threads the server tells the client to use. By caching the threads, we can avoid the overhead of thread allocation on subsequent queries.

These commits require additional manual testing so they have not been merged into #322 just yet.


src/main/java/com/ocient/jdbc/XGResultSet.java
```diff
@@ -86,64 +98,51 @@ public final class XGResultSet implements ResultSet
 				LOGGER.log(Level.WARNING, "Uncaught exception", throwable);
 			})
 			.build();
-		CONNECTION_REAPER = Executors.newSingleThreadExecutor(namedThreadFactory);
```
jwilliams-ocient, line 89: `CONNECTION_REAPER` was removed because there's no reason to offload the return-to-connection-pool work to a different thread.

jwilliams-ocient:
There's a bug lurking about here...
> Greg Dhuse
  [21 minutes ago](https://ocient.slack.com/archives/C04MV2RHXA4/p1679341834119119?thread_ts=1679319499.129489&cid=C04MV2RHXA4)
Looks like something about the new JDBC client (.105) caused the replay tool to hang (
[@Jordan Williams](https://ocient.slack.com/team/UT7ULG26A)
).  Maybe it waits for the client to close its connections, or maybe it hit some # connections limit?  Switching back to .104 seems to work, though

jwilliams-ocient:
This PR has been deprecated, new, non-thread based synchronization primitives have been added to #335

PR #325, opened 2023-03-21
wip[perf]: Non-blocking network I/O 

jwilliams-ocient:
## Overview
This PR lays the foundation for refactoring the networking architecture of the JDBC driver.

An overview of the changes can be found [here](https://ocient.atlassian.net/wiki/spaces/CON/pages/3735978097/JDBC+-+Resource+Management#Transport). 

### Transport Interface
The introduction of the `SimplexTransport` interface is the most significant change introduced in this pull request. 
https://github.com/Xeograph/ocient_jdbc/blob/ba49443cbadc0c4b5dff726c153a2fd56c727688/src/main/java/com/ocient/transport/SimplexTransport.java#L11-L30

This interface provides the following:
- [**perf**]: Increase client scalability. A non-blocking transport implementation will scale with the number of db connections better than a blocking implementation.
- [**refactor**]: Separate the wire format and transport implementations from the application logic. The refactor will make it significantly easier to implement additional wire formats.
- [**perf**]: Eliminate buffer copy from kernel to user space (java heap) when reading data off the network socket.
- [**perf**]: Fast-path to enable buffer pooling.

### JavaNetOcientWireV1Transport
Exists as a drop-in replacement for the existing `java.net` based implementation using the `SimplexTransport` interface
https://github.com/Xeograph/ocient_jdbc/blob/ba49443cbadc0c4b5dff726c153a2fd56c727688/src/main/java/com/ocient/transport/JavaNetOcientWireV1Transport.java#L33

### NettyOcientWireV1Transport
Provides a non-blocking wire transport implementation. On linux hosts, the client will prefer the edge-triggered `epoll` socket channel implementation over the level-triggered Java NIO one.
https://github.com/Xeograph/ocient_jdbc/blob/ba49443cbadc0c4b5dff726c153a2fd56c727688/src/main/java/com/ocient/transport/NettyOcientWireV1Transport.java#L56

### XGConnection
Factories for both transport implementations are provided 
https://github.com/Xeograph/ocient_jdbc/blob/ba49443cbadc0c4b5dff726c153a2fd56c727688/src/main/java/com/ocient/jdbc/XGConnection.java#L697-L707

## CompletableFuture
To fully realize the benefits of the non-blocking architecture, legacy code that previously operated directly on the network socket directly must be rewritten in an asynchronous manner. That is the bulk of the remaining work here.

### XGConnection
Helper methods have been added to `XGConnection` to help with sending requests over the transport interface. Prefer usage of these helper methods over using the `transportOrNull` member variable directly.
https://github.com/Xeograph/ocient_jdbc/blob/ba49443cbadc0c4b5dff726c153a2fd56c727688/src/main/java/com/ocient/jdbc/XGConnection.java#L5102-L5197

### ExceptionHandling
The pattern used for exception handling is to fast-fail on I/O submission failure. This is why `SimplexTransport#sendAndReceive` throws `IOException` and `InterruptedException`. When handling the result of a `CompletionStage`, great care must be taken to ensure error paths are handled correctly.

Prefer [CompletionStage#handle](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html#handle-java.util.function.BiFunction-) when the exceptional path should be handled. This is typically required when manually managing the lifecycle of some resource.

## Testing
I am going to add a unit testing infrastructure for XGConnection, since it's quite difficult to test this at the unit level.


src/main/java/com/ocient/transport/JavaNetOcientWireV1Transport.java
```diff
@@ -0,0 +1,301 @@
...
+            handleServerSignal(messageLength);
+            return StageSupport.exceptionalStage(new IOException("Handshake was rejected due to server signal"));
+        }
+
+        // Allocate a buffer to store the message
+        final ByteBuf byteBuf = allocate(messageLength);
+        try {
+            byte[] buf = byteBuf.array();
+            readBytes(buf, messageLength);
+            return StageSupport.completedStage(byteBuf);
```
jwilliams-ocient, line 203: we might need to manually advance the write index here, remind me to provide unit tests for this...

src/main/java/com/ocient/transport/SimplexTransportFactory.java
```diff
@@ -0,0 +1,20 @@
...
+     * Connects to a remote host using the specified transport and configuration.
+     *
+     * @param host the hostname or IP address of the remote host
+     * @param port the port number of the remote host
+     * @param tls the TLS configuration to use for the connection, or null if TLS is not required
+     * @param connectTimeout the maximum time to wait for the connection to be established, in milliseconds
+     * @return a new {@link SimplexTransport} object representing the connection to the remote host
+     * @throws Exception if an error occurs during the connection process
+     */
+    public T connect(final String host, final int port, Tls tls, final int connectTimeout) throws IOException;
```
jwilliams-ocient, line 19:
```suggestion
    public CompletionStage<T> connect(final String host, final int port, Tls tls, final int connectTimeout) throws IOException, InterruptedException;
```

jwilliams-ocient, line 19:
```suggestion
    public CompletionStage<T> connect(final String host, final int port, Tls tls, final int connectTimeout) throws IOException;
```

src/main/java/com/ocient/transport/JavaNetOcientWireV1Transport.java
jwilliams-ocient, line 203: we might need to manually advance the write index here, remind me to provide unit tests for this...

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -5087,192 +5014,186 @@ public <T> T unwrap(final Class<T> iface) throws SQLException
...
+								receivedSecurityToken.getIssuerFingerprint().toString()));
+					}
+				});
 	}
 
 	// Requests that a specific token be refreshed. In the event that the token is expired, but
 	// this connection has yet to establish a session, we cannot refresh the session. We can
 	// only refresh the token.
-	public Session.State sendRefreshToken(final Session.State oldState) throws SQLException{
+	public CompletionStage<Session.State> sendRefreshToken(final Session.State oldState) throws IOException, InterruptedException {
```
jwilliams-ocient, line 5060: This method was rewritten to provide an example of how to rewrite an existing synchronous method asynchronously.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4607,7 +4533,8 @@ private void sendSetSchema(final String schema) throws Exception
 			if(e instanceof SQLException && SQLStates.SESSION_EXPIRED.equals((SQLException) e)){
 				LOGGER.log(Level.INFO, "sendSetSchema() received session expired. Attempting to refresh session");
 				// Refresh my session.
-				refreshSession();
+				// FIXME this is now async
+				CompletableFutures.blockingGet(refreshSession().toCompletableFuture());
```
jwilliams-ocient, line 4537: Eventually, we will want to remove these blocking calls that were added.

PR #329, opened 2023-03-28
perf: only reset remote connection state when necessary

jwilliams-ocient:
This PR addresses an inefficiency in the connection pooling logic. The remote connection maintains some state that must be reset when activating a connection for a consumer with properties that differ from that of the consumer that recycled the connection.

If the activator's requested properties match, there is no work to be done. For whatever, reason, this logic was overly complex. Ideally, all of this connection property stuff if maintained separately from the XGConnection implementation but I didn't want to deal with scope creep in this respect.


jwilliams-ocient:
Local CI run:
```
[----------] 11 tests from OcientJDBCFixture (352.951s total)

[==========] 11 tests from 1 test case ran. (363.328s total)
[  PASSED  ] 11 tests.
```

PR #336, opened 2023-04-04
feat: rewrite XGConnection#reconnect and #sendParameter asynchronously [DB-23682] [DB-23683]

awang-ocient:
As the title suggests, I am refactoring reconnect() in XGConnection to be asynchronous.


src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -2450,7 +2451,14 @@ private int setParallelismSQL(final String cmd) throws SQLException
 		catch (final NumberFormatException e)
 		{
 		}
-		return conn.setParallelism(parallelism, reset);
+
```
awang-ocient, line 2454: I plan to change these blocking class to utilize `CompletableFutures.blockingGet()`

src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -2450,7 +2451,14 @@ private int setParallelismSQL(final String cmd) throws SQLException
...
+		// FIX ME
+		// function is asnyc now
+		try {
+			return conn.setParallelism(parallelism, reset).toCompletableFuture().get();
+		} catch (InterruptedException | ExecutionException e) {}
+
+		return 1;
```
awang-ocient, line 2461: Not sure what value to return here?

src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -2450,7 +2451,14 @@ private int setParallelismSQL(final String cmd) throws SQLException
 		catch (final NumberFormatException e)
 		{
 		}
-		return conn.setParallelism(parallelism, reset);
+
+		// FIX ME
+		// function is asnyc now
+		try {
+			return conn.setParallelism(parallelism, reset).toCompletableFuture().get();
```
awang-ocient, line 2458: I plan to change these blocking class to utilize `CompletableFutures.blockingGet()`

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4654,36 +4664,48 @@ private void sendClose() throws Exception
...
-					CompletableFutures.blockingGet(refreshSession().toCompletableFuture());
-				}
-				catch(Exception e) {}
-				// Now we should be able to re-run the command.
-				return sendParameterMessage(param);
+				// Refresh the now-expired session and then
+				// re-run the command
+				return refreshSession().thenCompose(v -> {
+					try {
+						return sendParameterMessage(param).thenApply(confirmationReponse -> 0);
```
jwilliams-ocient, line 4698: Recursion here can be tricky, there's a possibility we hit a stack overflow when futures complete on the calling thread. We may need to unroll this. Can you add a comment for me to look at this?
```suggestion
                                                FIXME Do we need to unroll the stack here? 
						return sendParameterMessage(param).thenApply(confirmationReponse -> 0);
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4654,36 +4664,48 @@ private void sendClose() throws Exception
...
 		final Request wrapper = builder.build();
 
-		try
-		{
-			out.write(intToBytes(wrapper.getSerializedSize()));
-			wrapper.writeTo(out);
-			out.flush();
-			getStandardResponse();
-		} catch (final Exception ex)
+		try {
```
jwilliams-ocient, line 4674: wow I think the Result API will clean up the exception handling nicely here...

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3976,32 +3975,39 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
+				// Connection superclass requires setSchema to return void
+				// so CompletionStage<Void> cannot be used here
 				setSchema(schema);
 			}
 			catch (final Exception e)
 			{
 			}
 		});
 
+		CompletionStage<Void> setPsoCompletionStage = null;
```
jwilliams-ocient, line 3995: (nit)

I typically avoid nulls if possible, and if something can legitimately be nullable, I'll leave an indicator in the variable name. In this case, the asynchronous execution does not produce a result so what you really want a way to say there is nothing to do.
```suggestion
		CompletionStage<Void> setPsoCompletionStage = StageSupport.voidStage();
```
This way, you can get rid of the null check and always return `setPsoCompletionStage`.

Just a little defensive programming trick.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3976,32 +3975,39 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
+	// Blocking IO
+	private CompletionStage<Void> clientHandshakeFinalize() throws SQLException, Exception {
 		// Set the default schema on the connection if one has not already been set
 		resetSchema().ifPresent(schema -> {
 			try
 			{
+				// Blocking IO
+				// Connection superclass requires setSchema to return void
+				// so CompletionStage<Void> cannot be used here
 				setSchema(schema);
```
jwilliams-ocient, line 3988: I'm glossing over the exceptional paths (sigh)

```suggestion
		return resetSchema()
		    .map(this::setSchemaAsync)
		    .orElseGet(StageSupport::voidStage)
		    .thenCompose(schemaUpdated -> {
		        if (setPso.isPresent()) {
 			        if (setPso.getAsLong() == -1)
 			        {
 				        // We have to turn it off
 				        return setPSO(false);
 			        }
 			        else if (setPso.getAsLong() > 0)
 			        {
 				        // Set non-default threshold
 				        return setPSO(setPso.getAsLong());
 			        }
 		        }
 		        
 		        // nothing more to do
 		        return StageSupport.voidStage();
		    });
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1455,6 +1455,25 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+			}
+		});
+
+		return clientHandshakeCompletionStage.thenRun(() -> {
+			try {
+				clientHandshakeFinalize();
+			} catch (Exception e) {
+				LOGGER.log(Level.WARNING, String.format("Handshake exception %s occurred in reconnect() with message %s", e.toString()));
+			}
+		});
```
jwilliams-ocient, line 1472: this method now returns a future, so you'd want to "chain" _this_ stage onto the completion of the previous stage `clientHandshakeCompletionStage` using composition, i.e. `thenCompose(...)`. 
```suggestion
		return clientHandshakeCompletionStage.thenCompose(this::clientHandshakeFinalize);
```

jwilliams-ocient (changes requested):
Mostly minor issues, happy to discuss via huddle zoom if that works best for you 👍🏾

awang-ocient:
> Mostly minor issues, happy to discuss via huddle zoom if that works best for you 👍🏾

I think I resolved all the comments

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -5088,17 +5119,27 @@ public void setSchema(final String schema) throws SQLException
...
 		try
 		{
 			// This could have been generated as a "cached" connection, which means it wasn't connected.
 			if(out == null){
-				reconnect();
+				completionStage = reconnect();
 			}						
-			sendSetSchema(schema);
+
+			completionStage = completionStage.thenCompose(v -> {
```
jwilliams-ocient, line 5131: There's a subtle distinction between the `thenCompose` and `thenComposeAsync` methods I'd like to take the opportunity to call attention to:

- `thenCompose()`: Returns a new CompletionStage that, when this stage completes normally, is executed with this stage as the argument to the supplied function.
- `thenComposeAsync()`: Returns a new CompletionStage that, when this stage completes normally, is executed using this stage's default asynchronous execution facility, with this stage as the argument to the supplied function.
 
In this case, we want to use the latter to avoid a deadlock situation. A general rule of thumb is to fire callbacks that execute synchronous, blocking code on a separate thread from the one on which the callback is being fired. 

This is extremely non-trivial stuff and dips into the finer implementation details of the `CompletableFuture` library, but it's a great exercise. Take a look when you get a chance! https://www.hackerrank.com/codepair/ymflrslcdzzgujhzkwvzhxvjbxaqzlwx/questions/1?b=eyJpbnRlcnZpZXdfaWQiOjQwNTMwMTQsInJvbGUiOiJpbnRlcnZpZXdlciIsInNob3J0X3VybCI6Imh0dHBzOi8vaHIuZ3MvYTE0MmI0NyIsImNhbmRpZGF0ZV91cmwiOiJodHRwczovL2hyLmdzL2Q0MGI2NmYifQ

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -5043,14 +5072,17 @@ public int setPriorityAdjustTime(final Integer priorityAdjustTime, final boolean
...
-		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.	
-		if(reset){
-			this.priorityAdjustTime = null;
-		} else {
-			this.priorityAdjustTime = priorityAdjustTime;
-		}
-		return rowsModified;
+		CompletionStage<Integer> rowsModifiedCompletionStage = sendParameterMessage(builder.build());
+
+		return rowsModifiedCompletionStage.thenApply(rowsModified -> {
```
jwilliams-ocient, line 5077: (uber nit):

```suggestion
		return sendParameterMessage(builder.build()).thenApply(rowsModified -> {
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -5024,17 +5050,20 @@ public int setServiceClassName(final String serviceClassName, final boolean rese
...
-		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.	
-		if(reset){
-			this.serviceClassName = null;
-		} else {
-			this.serviceClassName = serviceClassName;
-		}
-		return rowsModified;
+		CompletionStage<Integer> rowsModifiedCompletionStage = sendParameterMessage(builder.build());
+
+		return rowsModifiedCompletionStage.thenApply(rowsModified -> {
```
jwilliams-ocient, line 5055: (uber nit):

```suggestion
		return sendParameterMessage(builder.build()).thenApply(rowsModified -> {
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -5005,17 +5028,20 @@ public int setPriorityAdjustFactor(final Double priorityAdjustFactor, final bool
...
-		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.	
-		if(reset){
-			this.priorityAdjustFactor = null;
-		} else {
-			this.priorityAdjustFactor = priorityAdjustFactor;
-		}
-		return rowsModified;
+		CompletionStage<Integer> rowsModifiedCompletionStage = sendParameterMessage(builder.build());
+
+		return rowsModifiedCompletionStage.thenApply(rowsModified -> {
```
jwilliams-ocient, line 5033: (uber nit): same here

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4984,19 +5006,20 @@ public void setPSO(final long threshold) throws Exception
...
 		try
 		{
-			out.write(intToBytes(wrapper.getSerializedSize()));
-			wrapper.writeTo(out);
-			out.flush();
-			getStandardResponse();
+			CompletionStage<ClientWireProtocol.ConfirmationResponse> completionStage = sendRequestWithStandardResponse(wrapper);
+
+			// Return CompletionStage<Void> on success
+			return completionStage.thenAccept(Response -> {});
```
jwilliams-ocient, line 5012: (nit):

Prefer `thenApply()` to convey mapping from `T` => `U`. In this case `U` happens to be `Void` so the net effect is mostly the same, but the intent is much clearer when using `thenApply(...)`.
```suggestion
			return sendRequestWithStandardResponse(wrapper).thenApply(ign -> null);
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4946,20 +4967,21 @@ public void setPSO(final boolean on) throws Exception
...
 		try
 		{
-			out.write(intToBytes(wrapper.getSerializedSize()));
-			wrapper.writeTo(out);
-			out.flush();
-			getStandardResponse();
+			CompletionStage<ClientWireProtocol.ConfirmationResponse> completionStage = sendRequestWithStandardResponse(wrapper);
+
+			// Return CompletionStage<Void> on success
+			return completionStage.thenAccept(response -> {});
```
jwilliams-ocient, line 4973: (nit)

same here RE `thenApply`

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4868,7 +4889,7 @@ public int setPriority(final Double priority, final boolean reset) throws SQLExc
 		innerBuilder.setPriority(priority != null ? priority : 0.0);
 		builder.setPriority(innerBuilder.build());
 
-		int rowsModified = sendParameterMessage(builder.build());
+		CompletionStage<Integer> rowsModified = sendParameterMessage(builder.build());
 		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.	
 		if(reset){
 			this.priority = null;
```
jwilliams-ocient, line 4895: The logic below is intended to be executed _after_ `sendParameterMessage` succeeds

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4828,15 +4849,15 @@ public int setMaxTempDisk(final Integer maxTempDisk, final boolean reset) throws
...
 		LOGGER.log(Level.INFO, String.format("Setting maxTime to: %d", maxTime));
 		final ClientWireProtocol.SetParameter.Builder builder = ClientWireProtocol.SetParameter.newBuilder();
 		builder.setReset(reset);
 		final ClientWireProtocol.SetParameter.TimeLimit.Builder innerBuilder = ClientWireProtocol.SetParameter.TimeLimit.newBuilder();
 		innerBuilder.setTimeLimit(maxTime != null ? maxTime : 0);
 		builder.setTimeLimit(innerBuilder.build());
-		int rowsModified = sendParameterMessage(builder.build());		
+		CompletionStage<Integer> rowsModified = sendParameterMessage(builder.build());		
 		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.
 		if(reset){
```
jwilliams-ocient, line 4862: same here, this needs to be done after `sendParameterMessage()` completes successfully

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4810,15 +4831,15 @@ public int setMaxRowsHardLimit(final Integer maxRows, final boolean reset) throw
...
 		LOGGER.log(Level.INFO, String.format("Setting maxTempDisk to: %d", maxTempDisk));
 		final ClientWireProtocol.SetParameter.Builder builder = ClientWireProtocol.SetParameter.newBuilder();
 		builder.setReset(reset);
 		final ClientWireProtocol.SetParameter.MaxTempDiskLimit.Builder innerBuilder = ClientWireProtocol.SetParameter.MaxTempDiskLimit.newBuilder();
 		innerBuilder.setTempDiskLimit(maxTempDisk != null ? maxTempDisk : 0);
 		builder.setTempDiskLimit(innerBuilder.build());
-		int rowsModified = sendParameterMessage(builder.build());	
+		CompletionStage<Integer> rowsModified = sendParameterMessage(builder.build());	
 		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.
 		if(reset){
```
jwilliams-ocient, line 4844: same here, this needs to be done after `sendParameterMessage()` completes successfully

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4800,7 +4821,7 @@ public int setMaxRowsHardLimit(final Integer maxRows, final boolean reset) throw
 		final ClientWireProtocol.SetParameter.RowLimit.Builder innerBuilder = ClientWireProtocol.SetParameter.RowLimit.newBuilder();
 		innerBuilder.setRowLimit(maxRows != null ? maxRows : 0);
 		builder.setRowLimit(innerBuilder.build());
-		int rowsModified = sendParameterMessage(builder.build());
+		CompletionStage<Integer> rowsModified = sendParameterMessage(builder.build());
 		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.
 		if(reset){
 			this.maxRows = null;
```
jwilliams-ocient, line 4827: same here, this needs to be done after `sendParameterMessage()` completes successfully

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4756,7 +4777,7 @@ public int setParallelism(final Integer parallelism, final boolean reset) throws
 		innerBuilder.setConcurrency(parallelism != null ? parallelism : 0);
 		builder.setConcurrency(innerBuilder.build());
 
-		int rowsModified = sendParameterMessage(builder.build());
+		CompletionStage<Integer> rowsModified = sendParameterMessage(builder.build());
 		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.
 		if(reset){
 			this.parallelism = null;
```
jwilliams-ocient, line 4783: same here, this needs to be done after `sendParameterMessage()` completes successfully

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4698,29 +4725,23 @@ private void sendSetSchema(final String schema) throws Exception
...
 		{
-			if(e instanceof SQLException && SQLStates.SESSION_EXPIRED.equals((SQLException) e)){
-				LOGGER.log(Level.INFO, "sendSetSchema() received session expired. Attempting to refresh session");
-				// Refresh my session.
-				// FIXME this is now async
-				CompletableFutures.blockingGet(refreshSession().toCompletableFuture());
-				// Now we should be able to re-run the command.
-				sendSetSchema(schema);
-				return;
-			}
```
jwilliams-ocient, line 4718: I think we still need this

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4698,29 +4725,23 @@ private void sendSetSchema(final String schema) throws Exception
...
 		try
 		{
-			out.write(intToBytes(wrapper.getSerializedSize()));
-			wrapper.writeTo(out);
-			out.flush();
-			getStandardResponse();
+			CompletionStage<ClientWireProtocol.ConfirmationResponse> response = sendRequestWithStandardResponse(wrapper);
+
+			// Return CompletionStage<Void> on success
+			completionStage = response.thenAccept(res -> {});
```
jwilliams-ocient, line 4735: (nit)

```suggestion
			return sendRequestWithStandardResponse(wrapper).thenApply(ign -> null);
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4654,39 +4668,52 @@ private void sendClose() throws Exception
...
+						br.mergeFrom(new ByteBufInputStream(wireMessage));
+						return br.build();
+					},
+					// Reponse is already of type ConfirmationResponse
+					// identity function is sufficient
+					id -> id
+				);
+
+			// Return zero on success
+			return completionStage.thenApply(confirmationReponse -> 0);
```
jwilliams-ocient, line 4692: I wonder why we return 0 here at all?

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4349,73 +4359,77 @@ private static void validateConnectionProperties(Properties properties) throws S
 	 * FIXME there is absolutely no reason at all this can't be accomplished in 1 RT...
 	 */
 
-	private void resendParameters()
+	private CompletionStage<Void> resendParameters()
 	{
 		LOGGER.log(Level.INFO, "resendParameters() called");
+		List<CompletionStage<?>> completionStages = new ArrayList<>();
```
jwilliams-ocient, line 4365: Unfortunately, this approach won't work because it fires param requests in parallel, when they actually need to be executed serially due to the simplex nature of the protocol. One day soon, I hope to be able to do this - full duplex protocol is on the roadmap, but likely not until like v25

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3976,32 +3976,42 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
-			}
-		}
-
-		resendParameters();
+		return resetSchema()
+		    .map(schema -> {
+				try {
+					return setSchemaAsync(schema);
+				} catch (SQLException e) {
+					LOGGER.log(Level.WARNING, String.format("Exception %s occurred in clientHandshakeFinalize() with message %s", e.toString(), e.getMessage()));
```
jwilliams-ocient, line 3989: Need to propagate the exception here

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3976,32 +3976,42 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
+			})
+		    .orElseGet(StageSupport::voidStage)
+		    .thenCompose(schemaUpdated -> {
+		        if (setPso.isPresent()) {
+ 			        if (setPso.getAsLong() == -1)
+ 			        {
+ 				        // We have to turn it off
+ 				        try {
+							return setPSO(false);
+						} catch (Exception e) {}
```
jwilliams-ocient, line 4001: need to propagate the exception here

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3976,32 +3976,42 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
+		    .map(schema -> {
+				try {
+					return setSchemaAsync(schema);
+				} catch (SQLException e) {
+					LOGGER.log(Level.WARNING, String.format("Exception %s occurred in clientHandshakeFinalize() with message %s", e.toString(), e.getMessage()));
+				}
+				return StageSupport.voidStage();
+			})
+		    .orElseGet(StageSupport::voidStage)
+		    .thenCompose(schemaUpdated -> {
```
jwilliams-ocient, line 3994: Might need to handle the error here

awang-ocient: Not sure which line you're referencing here?

src/main/java/com/ocient/jdbc/XGConnection.java
awang-ocient: I believe the previous synchronous version of the function also returned 0. TBH I didn't look into the reasoning. Maybe we can return `CompletionStage<Void>` instead?

src/main/java/com/ocient/jdbc/XGConnection.java
awang-ocient: The previous try block called `getStandardResponse` which throws Exception. The current try block calls `sendRequestWithStandardResponse` which throws IOException but not SQLException.

I don't think it will compile if we try to catch SQLException. Thoughts here? Should `sendRequestWithStandardResponse` throw SQLException?

src/main/java/com/ocient/jdbc/XGConnection.java
awang-ocient: I changed it to `thenComposeAsync` for now, but actually `sendSetSchema()` doesn't appear to be blocking? It returns CompletionStage<Void>

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4349,73 +4368,91 @@ private static void validateConnectionProperties(Properties properties) throws S
...
+				} catch (Exception e) {
+					// This should never happen. We protect against bad arguments for these settings in the driver default. Also, we protect
+					// against them when these methods are called from statement.executeUpdate(...). So the arguments here should always be valid.
+					LOGGER.log(Level.WARNING, String.format("resendParameters got unexpected exception %s with message %s", e.toString(), e.getMessage()));
+					return CompletableFuture.failedFuture(e);
+				}
+			});
+		}
+		
+		return allStages.thenAccept(r -> {});
```
jwilliams-ocient, line 4455: (nit)

This is super clever! Love this approach. 

This is a good opportunity to highlight the `AsyncIterator` interface which provides some syntactic sugar for something like this.

```suggestion
		// Supplier instead of Callable so we don't have to handle the exception on the synchronous interface
                return AsyncIterator
                    .toIterator(setParameterSuppliers)
                    .thenCompose(Supplier::get)
                    
                    // Returns a CompletionStage<List<SQLResult<Void>>> so the caller can decide what to do if any tasks failed
                    .collect(Collectors.toList());
```

src/main/java/com/ocient/jdbc/XGConnection.java
jwilliams-ocient: yea idk, either, maybe leave it and add TODO to figure out what the legacy impl was trying to do

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4654,39 +4691,52 @@ private void sendClose() throws Exception
...
+					},
+					// Reponse is already of type ConfirmationResponse
+					// identity function is sufficient
+					id -> id
+				);
+
+			// Return zero on success
+			return completionStage.thenApply(confirmationReponse -> 0);
+		}
+		catch (final Exception ex)
```
jwilliams-ocient, line 4717: So the `feature/nio` branch now removes this exceptional path from the `sendRequest(...)` interface. If the request fails exceptionally, the underlying `TransportException` in the `TransportResult<?>` will indicate whether the request failed on send or recv. Special recovery logic can be added for either case, if necessary.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4654,39 +4691,52 @@ private void sendClose() throws Exception
...
 			}				
+
 			LOGGER.log(Level.WARNING, String.format("Failed sending set parameter request to the server with exception %s with message %s", ex, ex.getMessage()));
 			throw SQLStates.newGenericException(ex);
 		}
-		return 0;
 	}
 
-	private void sendSetSchema(final String schema) throws Exception
+	private CompletionStage<Void> sendSetSchema(final String schema) throws Exception
```
jwilliams-ocient, line 4739: Can return a `SQLResult` now and remove the exception from the method signature
```suggestion
	private CompletionStage<SQLResult<Void>> sendSetSchema(final String schema)
```

src/main/java/com/ocient/jdbc/XGConnection.java
jwilliams-ocient: I see, so I think this is a common pattern in this package:
1. Send request to DB that may complete exceptionally with SQLException
2. If the response contains the `SESSION_EXPIRED_CODE`, refresh the session, then retry the operation.

We can add a `recoverAsync(...)` method to the `*Result` interfaces:
```
/**
 * Maps the error value of this SQLResult to a new completion stage that
 * completes with a new SQLResult by applying the given function. If this 
 * result represents a success, the returned stage will complete with
 * the result.
 *
 * @param fn the function to apply to the error value
 */
default CompletionStage<SQLResult<T>> recover(final Function<? super E, ? extends CompletionStage<SQLResult<T>>> fn) {
    return fold(
        SQLResult::successAsync, 
        fn::apply);
}
```

And then use it here like:
```
sendRequestWithStandardResponse(request)
  .recoverAsync(ex -> {
    if (ex.getErrorCode() == SQLStates.SESSION_EXPIRED.getErrorCode()) {
      return refreshSession()
        .thenCompose(successOrEx -> successOrEx
          .mapSuccessAsync(ign -> sendRequestWithStandardResponse(request)));
    }
    // else
    // propagate the original exception
    return SQLResult.failAsync(ex);
  });
```

src/main/java/com/ocient/jdbc/XGConnection.java
jwilliams-ocient:
```
// Returns a CompletionStage<List<SQLResult<Void>>> so the caller can decide what to do if any tasks failed
// i.e.
whateverThisMethodIs().thenApply(results -> {
  List<SQLResult<Void>> failedResults = results.stream().filter(SQLResult::isException).collect(Collectors.toList());
  if (failedResults.isEmpty()) {
    return SQLResult.voidResult();
  }
  // TODO add suppressed exceptions
  return failedResults.get(0);
});
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2881,9 +2900,11 @@ public void forceExternal(final boolean force) throws SQLException
 		innerBuilder.setIsOn(force);
 		builder.setForceExternal(innerBuilder.build());
 		builder.setReset(false);
-		sendParameterMessage(builder.build());	
+		CompletionStage<SQLResult<Integer>> completionStage = sendParameterMessage(builder.build());	
 		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.
 		forceExternal = force;
+
+		return completionStage.thenApply(respOrEx -> respOrEx.isException() ? SQLResult.fail(respOrEx.getException()) : SQLResult.voidResult());
```
jwilliams-ocient, line 2907: (nit)

This is a good use case for `SQLResult#mapSuccess`, converting `SQLResult<T>` => `SQLResult<U>` where the exception type remains the same. The `Functions::noop` method reference is just syntactic sugar.
```suggestion
		return completionStage.thenApply(respOrEx -> respOrEx.mapSuccess(Functions::noop));
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4345,39 +4424,41 @@ private void sendClose() throws Exception
...
-		} catch (final Exception ex)
-		{
-			if(ex instanceof SQLException && SQLStates.SESSION_EXPIRED.equals((SQLException) ex)){
-				LOGGER.log(Level.INFO, "sendParameterMessage() received session expired. Attempting to refresh session");
-				// Refresh the now-expired session.
-				// FIXME this is now async
-				try {
-					blockingGet(refreshSession().toCompletableFuture());
+		// Return zero on success
+		return sendRequestWithStandardResponse(wrapper, Functions::alwaysTrue)
```
jwilliams-ocient, line 4435: 🔥

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4345,39 +4424,41 @@ private void sendClose() throws Exception
...
-			}				
-			LOGGER.log(Level.WARNING, String.format("Failed sending set parameter request to the server with exception %s with message %s", ex, ex.getMessage()));
-			throw SQLStates.newGenericException(ex);
-		}
-		return 0;
+
+				// Otherwise propagate error
+				return SQLResult.failAsync(ex);
+			}))
+			.thenApply(respOrEx -> respOrEx.isSuccess() ? /* Return 0 on success */ SQLResult.success(0) : SQLResult.fail(respOrEx.getException()));
```
jwilliams-ocient, line 4458: (nit)

```suggestion
			.thenApply(respOrEx -> respOrEx.mapSuccess(constant(0))));
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4733,14 +4846,19 @@ public int setPriorityAdjustTime(final Integer priorityAdjustTime, final boolean
...
-		if(reset){
-			this.priorityAdjustTime = null;
-		} else {
-			this.priorityAdjustTime = priorityAdjustTime;
-		}
-		return rowsModified;
+		return sendParameterMessage(builder.build()).thenApply(rowsModifiedOrEx -> {
+			// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.	
+
+			if(rowsModifiedOrEx.isSuccess()) {
```
jwilliams-ocient, line 4852: 👍🏾

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4040,73 +4081,111 @@ private static void validateConnectionProperties(Properties properties) throws S
...
+			});
 		}
+		
+		return allStages;
+
+		// TODO
+		// Utilize Async Iterator
+		// return AsyncIterator
+		// 	.fromIterator(setParameterCalls.listIterator())
+		// 	.thenCompose(Supplier::get)
```
jwilliams-ocient, line 4185: Yep, totally fine to leave this as a TODO, I'm pretty sure your chaining implementation looks fine

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3489,39 +3512,30 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
-				{
-				}
+			return clientHandshakeAsync(user, pwd, database, shouldRequestVersion)
+					.thenCompose(successOrEx -> {
+						
+						if(successOrEx.isException()) {
+							try
+							{
+								LOGGER.log(Level.INFO, "[nio] XGConnection in.close");
+								closeTransport();
```
jwilliams-ocient, line 3522: This synchronous call should be issued in a separate thread than the callback via `thenComposeAsync(...)` OR by using the helper method `runAsyncSafe_FIXME(...)`.
```suggestion
								return runAsyncSafe_FIXME(this::closeTransport())
								    .thenApply(constant(successOrEx));
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3653,61 +3666,89 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
-			{
-			}
+		return clientHandshakeAsync(user, pwd, database, shouldRequestVersion)
+					.thenCompose(successOrEx -> {
+						
+						if(successOrEx.isException()) {
+							try
+							{
+								LOGGER.log(Level.INFO, "[nio] XGConnection in.close");
+								closeTransport();
```
jwilliams-ocient, line 3676: Same here RE: `runAsyncSafe_FIXME(...)`

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3653,61 +3666,89 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
-			{
-				// Set non-default threshold
-				setPSO(setPso.getAsLong());
-			}
-		}
+		return resetSchema()
+		    .map(schema -> {
+				try {
+					return setSchemaAsync(schema);
+				} catch (SQLException e) {
```
jwilliams-ocient, line 3704: Should be able to remove this catch block. There's a quite a bit of duplicated log messages in the legacy implementation, I think this is an example of one that can be removed since the caller of `clientHandshakeFinalize` will surely handle the error.

If the log message provided some additional context i.e. failed to set the default schema, iit could provide value but doesn't as is.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4605,7 +4701,7 @@ public void setPSOSeed(final long seed, boolean reset) throws Exception
 	}
 	// sets the pso threshold on this connection to be -1(meaning pso is turned off)
 	// or back to the default
-	public void setPSO(final boolean on) throws Exception
+	public CompletionStage<SQLResult<Void>> setPSO(final boolean on) throws SQLException
```
jwilliams-ocient, line 4704:
```suggestion
	public CompletionStage<SQLResult<Void>> setPSO(final boolean on)
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4635,21 +4731,30 @@ public void setPSO(final boolean on) throws Exception
...
+		// catch (final SQLException e) {
+		// 	// Doesn't matter...
+		// 	LOGGER.log(Level.WARNING, String.format("Failed sending set pso request to the server with exception %s with message ", e.toString(), e.getMessage()));
+		// 	throw e;
+		// }
 	}
 
 	// sets the pso threshold on this connection to threshold
-	public void setPSO(final long threshold) throws Exception
+	public CompletionStage<SQLResult<Void>> setPSO(final long threshold) throws SQLException
```
jwilliams-ocient, line 4757:
```suggestion
	public CompletionStage<SQLResult<Void>> setPSO(final long threshold)
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4672,21 +4777,19 @@ public void setPSO(final long threshold) throws Exception
...
+				LOGGER.log(Level.WARNING, String.format("Failed sending set pso request to the server with exception %s with message ", e.toString(), e.getMessage()));
+				return SQLResult.fail(e);
+			}
+
+			return SQLResult.voidResult();
+		});
 	}
 
-	public int setPriorityAdjustFactor(final Double priorityAdjustFactor, final boolean reset) throws SQLException
+	public CompletionStage<SQLResult<Integer>> setPriorityAdjustFactor(final Double priorityAdjustFactor, final boolean reset) throws SQLException
```
jwilliams-ocient, line 4792:
```suggestion
	public CompletionStage<SQLResult<Integer>> setPriorityAdjustFactor(final Double priorityAdjustFactor, final boolean reset)
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4768,31 +4886,26 @@ public Savepoint setSavepoint(final String arg0) throws SQLException
...
-		{
-			sendSetSchema(schema);
-		}
-		catch (final Exception e)
-		{
+		return sendSetSchema(schema);
+	}
+
+	@Override
+	public void setSchema(final String schema) throws SQLException
```
jwilliams-ocient, line 4902: `SQLException` stays on this synchronous interface from `java.sql`.

awang-ocient:
`reconnect()` has a chunk of code which tries to connect to all secondary interfaces in a serial fashion.

https://github.com/Xeograph/ocient_jdbc/blob/4602079730d88ce14f5dcba0e4351031162f575b/src/main/java/com/ocient/jdbc/XGConnection.java#L3542-L3569

The old logic was that if the connect to an interface failed (throws an Exception), then the for loop would continue to the next iteration and try the next interface.

However, because of our async refactor, `clientHandshakeAsync()` no longer throws an Exception, so I believe the for loop terminates after the first iteration because the return statement executes...

How can we work around this? It seems like a different issue because `connect` and `clientHandshakeAsync()` are logically related, but `connect` is blocking whereas `clientHandshakeAsync()` is non-blocking.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1358,6 +1361,22 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+			try { clientHandshake(user, pwd, database, shouldRequestVersion); }
+			catch (Throwable t) {
+				LOGGER.log(Level.WARNING, String.format("Handshake exception %s occurred in reconnect() with message %s", t.toString()));
+
+				// wrap the throwable in a completion exception so it can be handled by the caller
+				throw new CompletionException(t);
+			}
+		});
+
+		return clientHandshakeCompletionStage.thenCompose(v -> clientHandshakeFinalize());
```
jwilliams-ocient, line 1375: If `clientHandshake(...)` completes exceptionally, I think the returned `CompletionStage<SQLResult<Void>>` will also complete exceptionally, which we actually don't want here. We want to propagate the exception through the `SQLResult` container.
```suggestion
                return runAsync_FIXME(() -> clientHandshake(user, pwd, database, shouldRequestVersion))
                    .thenCompose(successOrEx -> successOrEx
                        .ampSuccessAsync(ign -> clientHandshakeFinalize()));
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2881,9 +2900,11 @@ public void forceExternal(final boolean force) throws SQLException
 		innerBuilder.setIsOn(force);
 		builder.setForceExternal(innerBuilder.build());
 		builder.setReset(false);
-		sendParameterMessage(builder.build());	
+		CompletionStage<SQLResult<Integer>> completionStage = sendParameterMessage(builder.build());	
 		// Change this only if sendParameterMessage succeeded. It would have thrown otherwise.
 		forceExternal = force;
```
jwilliams-ocient, line 2905: Based on the comment above, I think this should be done in the `completionStage.thenApply(...)` block below iff `respOrEx` represents a successful response.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3552,32 +3563,27 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
-						retVal = (SQLException) handshakeException;
-						LOGGER.log(Level.WARNING, String.format("Handshake exception %s occurred in reconnect() with message %s", retVal.toString(), retVal.getMessage()));
-					}
-				}
-				// reconnect failed so we are no longer connected
-				connected = false;
+							if(successOrEx.getException() instanceof SQLException) {
+								SQLException e = successOrEx.getException();
+								LOGGER.log(Level.WARNING, String.format("Handshake exception %s occurred in reconnect() with message %s", e.toString(), e.getMessage()));
+							}
```
jwilliams-ocient, line 3574: (nit)

I think you can remove this check since `successOrEx.getException() instanceof SQLException` will always be true because `SQLResult<T>` is a specialization of `Result<T, SQLException>`.
```suggestion
```

Then you could condense the parent block into:
```java
return clientHandshakeAsync(user, pwd, database, shouldRequestVersion)
    .thenCompose(successOrEx -> successOrEx
       .recoverAsync(ex -> runAsyncSafe_FIXME(this::closeTransport)
           .thenApply(whoCares -> {
             connected = false;
             return successOrEx;
           })));
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3606,10 +3612,10 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
 
 				try
 				{
-					clientHandshake(user, pwd, database, shouldRequestVersion);
-					clientHandshakeFinalize();
+					CompletionStage<SQLResult<Void>> completionStage = clientHandshakeAsync(user, pwd, database, shouldRequestVersion);
+
 					secondaryIndex = index;
-					return;
+					return completionStage;
```
jwilliams-ocient, line 3618: Does this return change the behavior of this function to fail-fast on the first failed handshake?

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3606,10 +3612,10 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
 
 				try
 				{
-					clientHandshake(user, pwd, database, shouldRequestVersion);
-					clientHandshakeFinalize();
+					CompletionStage<SQLResult<Void>> completionStage = clientHandshakeAsync(user, pwd, database, shouldRequestVersion);
+
 					secondaryIndex = index;
```
jwilliams-ocient, line 3617: probably want to do this after the future completes

src/main/java/com/ocient/jdbc/XGConnection.java
jwilliams-ocient: This will be _very_ hard to write in its current form due to the synchronous control mechanisms (for-loop and exception handling) used by this function.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3653,61 +3658,79 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
+		    .thenCompose(schemaUpdated -> {
+		        if (setPso.isPresent()) {
+ 			        if (setPso.getAsLong() == -1)
+ 			        {
+ 				        // We have to turn it off
+ 				        try {
+							return setPSO(false);
+						} catch (Exception e) {
+							return CompletableFuture.failedFuture(e);
+						}
```
jwilliams-ocient, line 3701:
```suggestion
 				        return setPSO(false);
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3653,61 +3658,79 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
+ 			        }
+ 			        else if (setPso.getAsLong() > 0)
+ 			        {
+ 				        // Set non-default threshold
+ 				        try {
+							return setPSO(setPso.getAsLong());
+						} catch (Exception e) {
+							return CompletableFuture.failedFuture(e);
+						}
```
jwilliams-ocient, line 3710:
```suggestion
 				        return setPSO(setPso.getAsLong());
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3653,61 +3658,79 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
-		resendParameters();
+ 		        // nothing more to do
+ 		        return SQLResult.voidFuture();
+		    })
+			.thenCompose(successOrEx -> { 
+
+				if(successOrEx.isException()) {
+					SQLException e = successOrEx.getException();
+					LOGGER.log(Level.WARNING, String.format("Exception %s occurred in clientHandshakeFinalize() with message %s", e.toString(), e.getMessage()));
+				}
```
jwilliams-ocient, line 3722: wondering if we actually want to swallow this exception, I'll have to look at the legacy impl. Sorry we don't have existing unit tests for stuff like this 🤦🏾

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3653,61 +3658,79 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
+
+				return resendParameters(); 
+			})
+			.thenApply(results -> {
+				List<SQLResult<Void>> failedResults = results.stream().filter(SQLResult::isException).collect(Collectors.toList());
+				if (failedResults.isEmpty()) {
+				  return SQLResult.voidResult();
+				}
+				// TODO add suppressed exceptions
+				return failedResults.get(0);
```
jwilliams-ocient, line 3732: 👍🏾

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4345,39 +4406,41 @@ private void sendClose() throws Exception
...
+					// Refresh session and try again
+					return sendRefreshSession()
+						.thenCompose(successOrEx -> successOrEx
+						.mapSuccessAsync(ign -> sendRequestWithStandardResponse(wrapper, Functions::alwaysTrue)
+							.thenApply(retryOrEx -> {
+
+								if(retryOrEx.isException()) {
+									// Retry (after refreshing the session) failed
+									SQLException e = retryOrEx.getException();
+									LOGGER.log(Level.WARNING, String.format("Failed sending set parameter request to the server after refresh with exception %s with message %s", e, e.getMessage()));
```
jwilliams-ocient, line 4430: (uber nit)

I've been making an effort to clean up some of the logging as I refactor (I know you didn't write it this way).
```suggestion
									LOGGER.log(Level.WARNING, retryOrEx.getException(), () -> String.format("Failed sending set parameter request to the server after refresh: %s, this));
```
Two reasons to do this:
1. perf: only materialize messages if the log level is enabled
2. functions: using the `java.util.logging` mechanism for logging exceptions means that we actually get stack traces. Typically, the message provides little to no value without the stack trace. It's unfortunate that someone that didn't have experience writing java architected the JDBC driver as we know it, which is why we're going through all of these pains now lol. Ideally, we'd be using log4j like any sane, modern-day library, but here we are. I actually have a story to do this work we might be able to squeeze in the next month... https://ocient.atlassian.net/browse/DB-22051

Clean up these instances, if you happen to already be touching the surrounding code. Another really bad habit I see is using `Level.INFO` for every fricken log message which effectively eliminates any possibility of filtering logs at the logging config level, which is just so ridiculous. Messages like "Called XYX" are clear ccandidates for "TRACE" level logging (or "FINEST" in the `java.util.lgging` world).

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4389,29 +4452,19 @@ private void sendSetSchema(final String schema) throws Exception
...
-			// Doesn't matter...
-			LOGGER.log(Level.WARNING, String.format("Failed sending set schema request to the server with exception %s with message %s", e.toString(), e.getMessage()));
-		}
+		return sendRequestWithStandardResponse(wrapper, Functions::alwaysTrue).thenApply(resOrEx-> {
+			setSchema = Optional.of(schema);
 
-		setSchema = Optional.of(schema);
+			if(resOrEx.isException()) {
+				SQLException e = resOrEx.getException();
+				LOGGER.log(Level.WARNING, String.format("Failed sending set schema request to the server with exception %s with message %s", e.toString(), e.getMessage()));
```
jwilliams-ocient, line 4460: we can probably kill this log message, just let the caller decide if they want to log the exception - they probably already do.

src/main/java/com/ocient/util/Functions.java
```diff
@@ -18,7 +18,7 @@ public static void noop() {}
      * @param t the argument to ignore
      * @param <T> the type of the argument
      */
-    public static <T> void noop(T t) {}
+    public static <T> Void noop(T t) { return null; }
```
jwilliams-ocient, line 21: I think you should add a new method for this as the no-return variant is surely used elsewhere in the codebase

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -179,6 +181,8 @@
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufInputStream;
 import io.netty.buffer.ByteBufOutputStream;
+import io.netty.channel.VoidChannelPromise;
```
jwilliams-ocient, line 184: can probably delete this import, prefer `*Result.voidFuture()` or `StageSupport.voidFuture()` for a non-exceptional future

src/main/java/com/ocient/jdbc/XGConnection.java
jwilliams-ocient: Just saw your comment above, you are spot on 👍🏾

jwilliams-ocient: I recommend decomposing this method into its logical parts:
1. Find _all_ known secondary interfaces paired with their relative index into the `secondaryInterfaces` list. Each secondary interface represents a database endpoint (annoying that we create obscure names for trivial, widely agreed upon concepts but w/e). If the client was connected to an endpoint at _some_ point in time (i.e. `secondaryIndex != -1`), the endpoint should be the first element in the returned list.
2. Iterate, asynchronously, over the list of known db endpoints, mapping each endpoint to a "connect-and-complete-handshake" routine that produces a `CompletionStage<SQLResult<Integer>>` that completes with a result representing the index of the endpoint in the `secondaryInterfaces` list upon success, else failure.
3. Your iterator should "find" the first endpoint that returns success, skipping endpoints that fail to connect.

I'd then recommend using the `AsyncIterator` interface to handle (2) and (3). See https://github.com/IBM/java-async-util/blob/0da9d9cb703ba3fd255c185d06aefa5f65de03eb/asyncutil/src/test/java/com/ibm/asyncutil/iteration/AsyncIteratorTest.java#L609-L615 for an example on how to use the "find" feature.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4040,73 +4064,111 @@ private static void validateConnectionProperties(Properties properties) throws S
...
+							resultList.add(SQLResult.fail(((SQLResult<?>) successOrEx).getException()));
+						}
+
+						return StageSupport.completedStage(resultList);
+					});
+				} catch (Exception e) {
+					// This should never happen. We protect against bad arguments for these settings in the driver default. Also, we protect
+					// 	against them when these methods are called from statement.executeUpdate(...). So the arguments here should always be valid.
+					LOGGER.log(Level.WARNING, String.format("resendParameters got unexpected exception %s with message %s", e.toString(), e.getMessage()));
+					return StageSupport.completedStage(resultList);
```
jwilliams-ocient, line 4156: (nit)

If you replace `Callable<T>` with `Supplier<T>` you can get rid of the try-catch block

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4345,39 +4407,41 @@ private void sendClose() throws Exception
...
+					// Refresh session and try again
+					return sendRefreshSession()
+						.thenCompose(successOrEx -> successOrEx
+						.mapSuccessAsync(ign -> sendRequestWithStandardResponse(wrapper, Functions::alwaysTrue)
+							.thenApply(retryOrEx -> {
+
+								if(retryOrEx.isException()) {
+									// Retry (after refreshing the session) failed
+									SQLException e = retryOrEx.getException();
+									LOGGER.log(Level.WARNING, e, () -> String.format("Failed sending set parameter request to the server after refresh with exception %s with message %s", e, e.getMessage()));
```
jwilliams-ocient, line 4431: (nit)

Avoid printing the message which will shown up in the stack trace and include `this` which has a `toString()` method that will let us know which channel the error happened on.
```suggestion
									LOGGER.log(Level.WARNING, e, () -> String.format("Failed sending set parameter request to the server after refresh: %s", this));
```

There's more I want to do here w.r.t. contextual logging. There's  pattern referred to as `MDC` in logging utilities that allows you to define thread local variables to provide additional context to log messages implicitly i.e. a query ID used to prefix each message. The pattern is trickier to implement in a non-blocking world where code can run an any thread so I didn't do that, be we should consider doing something like that in the future.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4389,29 +4453,16 @@ private void sendSetSchema(final String schema) throws Exception
...
+		return sendRequestWithStandardResponse(wrapper, Functions::alwaysTrue).thenApply(resOrEx-> {
+			setSchema = Optional.of(schema);
+
+			if(resOrEx.isException()) {
+				// Propagate exception to caller
+				return SQLResult.fail(resOrEx.getException());
+			}
 
-		setSchema = Optional.of(schema);
+			return SQLResult.voidResult();
```
jwilliams-ocient, line 4464: (uber nit)

```suggestion
			return resOrEx.mapSuccess(Functions::noopBoxedVoid);
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4768,31 +4857,24 @@ public Savepoint setSavepoint(final String arg0) throws SQLException
...
-			if (e instanceof SQLException)
-			{
-				throw (SQLException) e;
-			}
-			else
-			{
-				throw SQLStates.newGenericException(e);
-			}
+			throw new CompletionException(e);
 		}
```
jwilliams-ocient, line 4878:
```suggestion
		SQLResult<Void> successOrEx = blockingGetSQLResult(setSchemaAsync(schema));
        if (successOrEx.isException()) {
            LOGGER.log(Level.WARNING, successOrEx.getException(), () -> String.format("Failed to update schema: %s", this));
            throw successOrEx.getException();
        }
```

jwilliams-ocient (commented):
Awesome, should merge this afternoon. 1 issue, the rest are nits :smile:

jwilliams-ocient (commented):
I like to run these two test fixtures when making connection-related changes to the driver:

```shell
bazel/run_ci_fixture.py quiesce_round_robin --config test_release -- --ocient_jdbc <path_to_your_custom_jar>
```
and
```shell
bazel/run_ci_fixture.py full_stack_connection_minimal --config test_release -- --ocient_jdbc <path_to_your_custom_jar>
```

Might be a good idea to run these before merging too.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4040,73 +4063,103 @@ private static void validateConnectionProperties(Properties properties) throws S
 	 * FIXME there is absolutely no reason at all this can't be accomplished in 1 RT...
 	 */
 
-	private void resendParameters()
+	private CompletionStage<List<SQLResult<Void>>> resendParameters()
```
jwilliams-ocient, line 4066: So I get this error when simulating quiesce manually on your branch:

```
2023-04-27T12:50:13.071 [32] com.ocient.jdbc.XGStatement handleSendAndReceiveFailure WARNING: Error executing EXECUTE_QUERY request, failMode=RECV, channel=XGConnection[id=3, info=XGConnectionInfo[hint=STATEMENT, user='admin@system', pwd='*****', ip='localhost', portNum=4052, database='test', protocolVersion='7.1.0', tls=OFF, hints=[force=false, oneShotForce=false]], connected=true, socket=JavaNetOcientWireV1Transport [remote=localhost/127.0.0.1:4052, local=/127.0.0.1:38664]] 
java.sql.SQLException: 
	at com.ocient.jdbc.SQLStates.newGenericException(SQLStates.java:711)
	at com.ocient.jdbc.XGStatement.lambda$beginSendAndReceive$36(XGStatement.java:2373)
	at com.ocient.jdbc.XGConnection.lambda$sendRequest$125(XGConnection.java:5141)
	at com.ocient.util.CompletableFutures$SuccessResult.fold(CompletableFutures.java:226)
	at com.ocient.jdbc.XGConnection.lambda$sendRequest$126(XGConnection.java:5139)
	at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:646)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1773)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at com.ocient.jdbc.XGStatement.lambda$beginSendAndReceive$36(XGStatement.java:2369)
	... 12 more
Caused by: com.google.protobuf.InvalidProtocolBufferException: Protocol message contained an invalid tag (zero).
	at com.google.protobuf.InvalidProtocolBufferException.invalidTag(InvalidProtocolBufferException.java:102)
	at com.google.protobuf.CodedInputStream$StreamDecoder.readTag(CodedInputStream.java:2068)
	at com.ocient.jdbc.proto.ClientWireProtocol$ExecuteQueryResponse.<init>(ClientWireProtocol.java:38304)
	at com.ocient.jdbc.proto.ClientWireProtocol$ExecuteQueryResponse.<init>(ClientWireProtocol.java:38265)
	at com.ocient.jdbc.proto.ClientWireProtocol$ExecuteQueryResponse$1.parsePartialFrom(ClientWireProtocol.java:39373)
	at com.ocient.jdbc.proto.ClientWireProtocol$ExecuteQueryResponse$1.parsePartialFrom(ClientWireProtocol.java:39367)
	at com.ocient.jdbc.proto.ClientWireProtocol$ExecuteQueryResponse$Builder.mergeFrom(ClientWireProtocol.java:38921)
	at com.ocient.jdbc.proto.ClientWireProtocol$ExecuteQueryResponse$Builder.mergeFrom(ClientWireProtocol.java:38749)
	at com.google.protobuf.AbstractMessage$Builder.mergeFrom(AbstractMessage.java:420)
	at com.google.protobuf.AbstractMessage$Builder.mergeFrom(AbstractMessage.java:317)
	at com.google.protobuf.AbstractMessageLite$Builder.mergeFrom(AbstractMessageLite.java:260)
	at com.google.protobuf.AbstractMessage$Builder.mergeFrom(AbstractMessage.java:563)
	... 17 more
```

The "invalid tag" error means the application read garbage data off the socket. This is a little easier to debug because we know the protocol is simplex. The implication is that the client is sending requests concurrently. The "netty" impl has a proper guard for this (see the usage of the `AsyncLock` in that impl).

We have to make sure that the callers of this function are also waiting for this coroutine to complete before continuing.

jwilliams-ocient: It looks like `reset()` is at least one of the offenders

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -219,7 +219,7 @@ public void run()
 						throw e;
 					}
 
-					reconnectOrThrow();
+					blockingGetSQLResult(reconnect());
```
jwilliams-ocient, line 222: hmm, I'm not sure we want to swallow reconnect failures, I think what you had before was right...

Or `blockingGetSQLResult(reconnect()).get();` which is just syntactic sugar for your `reconnectOrThrow()` helper method.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -855,10 +857,17 @@ public CompletionStage<Void> refreshSession() {
...
 			.refresh(this.sessionState, this, false)
-			.thenAccept(newStateOrEx -> newStateOrEx.left().ifPresent(newState -> this.sessionState = newState));
+			.thenApply(newStateOrEx -> {
+				if(newStateOrEx.isException()) {
+					return SQLResult.fail(newStateOrEx.getException());
+				}
+
+				this.sessionState = newStateOrEx.left().get();
+				return SQLResult.voidResult();
+			});
```
jwilliams-ocient, line 870: (nit)

```suggestion
			.thenApply(newStateOrEx -> newStateOrEx
			    .mapSuccess(newState -> {
			        this.sessionState = newState;
			        return (Void) null;
			    }));
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3073,7 +3104,7 @@ private String getSchemaFromServer() throws Exception
 			}
 			else
 			{
-				reconnect();
+				blockingGetSQLResult(reconnect());
```
jwilliams-ocient, line 3107: I don't think we should swallow this exception
```suggestion
				blockingGetSQLResult(reconnect()).get();
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3853,15 +3862,15 @@ private void redirectInternal(final String host, final int port, final boolean s
 			{
 
 				LOGGER.log(Level.WARNING, String.format("Exception %s occurred in redirect() with message %s", e.toString(), e.getMessage()));
-				reconnect();
+				blockingGetSQLResult(reconnect());
```
jwilliams-ocient, line 3865:
```suggestion
				blockingGetSQLResult(reconnect()).get();
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3853,15 +3862,15 @@ private void redirectInternal(final String host, final int port, final boolean s
...
 				return;
 			}
 
 			try
 			{
 				clientHandshake(user, pwd, database, shouldRequestVersion);
 				oneShotForce = true;
-				clientHandshakeFinalize();
+				blockingGetSQLResult(clientHandshakeFinalize());
```
jwilliams-ocient, line 3873:
```suggestion
				blockingGetSQLResult(clientHandshakeFinalize()).get();
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3874,7 +3883,7 @@ private void redirectInternal(final String host, final int port, final boolean s
 				{
 				}
 
-				reconnect();
+				blockingGetSQLResult(reconnect());
```
jwilliams-ocient, line 3886:
```suggestion
				blockingGetSQLResult(reconnect()).get();
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4150,22 +4188,26 @@ void reset() throws SQLException
 					if (setPso.getAsLong() == -1)
 					{
 						// We have to turn it off
-						setPSO(false);
+						blockingGetSQLResult(setPSO(false));
```
jwilliams-ocient, line 4191:
```suggestion
						blockingGetSQLResult(setPSO(false)).get();
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4150,22 +4188,26 @@ void reset() throws SQLException
...
 					}
 					else if (setPso.getAsLong() > 0)
 					{
 						// Set non-default threshold
-						setPSO(setPso.getAsLong());
+						blockingGetSQLResult(setPSO(setPso.getAsLong()));
```
jwilliams-ocient, line 4196:
```suggestion
						blockingGetSQLResult(setPSO(setPso.getAsLong())).get();
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4150,22 +4188,26 @@ void reset() throws SQLException
...
 					} else {
-						setPSO(true);
+						blockingGetSQLResult(setPSO(true));
```
jwilliams-ocient, line 4198:
```suggestion
						blockingGetSQLResult(setPSO(true)).get();
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -4150,22 +4188,26 @@ void reset() throws SQLException
...
 					}
 				}
 			}
 			catch (final Exception e)
 			{
 			}
 			
-			resendParameters();
+			try {
+				blockingGet(resendParameters());
```
jwilliams-ocient, line 4207:
```suggestion
				blockingGet(resendParameters()).get();
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -5220,11 +5276,12 @@ public void activateObject(XGConnectionInfo info, PooledObject<XGConnection> poo
 						ign -> SQLResult.voidFuture(), 
 						recoverableEx -> {
 							if (conn.fatalException.get() != null) {
+								LOGGER.log(Level.INFO, "Found fatal exception, do not attempt to reconnect");
 								return SQLResult.failAsync(recoverableEx);
 							}
-
+							
 							LOGGER.log(Level.WARNING, "Connection dropped while idle, reconnecting", recoverableEx);
-							return runAsyncSafe_FIXME(conn::reconnect);
```
jwilliams-ocient, line 5227: 👍🏾

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -5265,18 +5322,25 @@ public PooledObject<XGConnection> makeObject(XGConnectionInfo info) throws Excep
 								conn.setFatalException(exception);
 								return SQLResult.failAsync(exception);
 							}
-							
+
 							LOGGER.log(Level.INFO, exception,
 									() -> String.format("Attempting reconnect after connection failure: %s", conn));
 
-							return runAsyncSafe_FIXME(conn::reconnect)
```
jwilliams-ocient, line 5272: 👍🏾

src/main/java/com/ocient/jdbc/XGResultSet.java
```diff
@@ -1884,39 +1884,42 @@ public CompletionStage<SQLResult<Void>> closeAsync() {
...
-							if ((e instanceof SQLException)) {
-								final SQLException sqlEx = ((SQLException) e);
-								final boolean channelIsReadyForNextQuery = 
-									sqlEx.getErrorCode() == SQLStates.OPERATION_CANCELED.getSqlCode() ||
-									sqlEx.getMessage().toLowerCase().contains("query was not found");
-								if (channelIsReadyForNextQuery) {
+						
+						boolean shouldRetryKillQuery = shouldKillQuery;
+						return m_conn
+							.reconnect()
```
jwilliams-ocient, line 1890: 👍🏾 ah another silent merge conflict I missed 😢

src/main/java/com/ocient/jdbc/XGStatement.java
```diff
@@ -2142,7 +2143,8 @@ public void passUpCancel(final boolean clearCancelFlag) throws SQLException
 
 	private void reconnect() throws IOException, SQLException
 	{
-		conn.reconnect();
+		// FIXME 
+		blockingGetSQLResult(conn.reconnect());
```
jwilliams-ocient, line 2147:
```suggestion
		blockingGetSQLResult(conn.reconnect()).get();
```

jwilliams-ocient (commented):
It looks like we're swallowing some exceptions. Is that intentional?

PR #340, opened 2023-04-11
fix: defer invalidating pooled objects for redirected connections unt…

jwilliams-ocient:
…il close

 * Invoking POOL#invalidate() with the pooled object will *immediately* trigger connection teardown, when we really just want to defer this to the end of the connection's lifecycle (i.e. end user calls XGConnection#close()"


awang-ocient (approved):
LGTM!

PR #342, opened 2023-04-17
perf: send heartbeat iff the connection has been idle for longer than…

jwilliams-ocient:
… 60s

 * chore: update version to v2.106.rc1


src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -5494,26 +5495,30 @@ public void activateObject(XGConnectionInfo info, PooledObject<XGConnection> poo
...
-				.thenAcceptAsync(isConnected -> {
-					// The callback is guaranteed to be invoked on an activator thread
-					if (!isConnected) {
-						try {
-							LOGGER.log(Level.WARNING, "Connection dropped while idle, reconnecting");
-							conn.reconnect();
-						} catch (Throwable t) {
-							conn.setFatalException(t);
-							throw new CompletionException(t);
+			if (System.currentTimeMillis() - conn.lastIdleTime > 60 * 1000) {
```
awang-ocient, line 5501: Maybe use a variable to store the idle value of 60s.

awang-ocient (approved):
LGTM

src/main/java/com/ocient/jdbc/XGConnection.java
jwilliams-ocient: Agreed, much more readable

PR #344, opened 2023-04-21
feat: Rewrite XGConnection#clientConnection* methods asynchronously [DB-23681]

awang-ocient:


jwilliams-ocient (commented):
ugh, I forgot to hit "Submit review"

jwilliams-ocient:
This looks phenomenal btw. Thanks for cleaning up some of the crypto stuff 🙏🏾

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1358,626 +1362,716 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+					LOGGER.log(Level.INFO, e, () -> "No Okta Authenticators found");
+					return SQLResult.failAsync(e);
+				}
+				
+				return clientHandshakeOktaSessionToken(db, authenticator.get(), shouldRequestVersion);
+			} else {
+				// GCM
+				return clientHandshakeGCM(userid, pwd, db, shouldRequestVersion, handshake);
 			}
+		} catch (Exception e) {
```
jwilliams-ocient, line 1470: Can we get rid of the try/catch blocks? Ideally, none of these methods throw on the synchronous interface

jwilliams-ocient: yea, it looks like you did remove the `throws` from the clientHandshake methods

jwilliams-ocient: ah it's the redirect which is still synchronous on `feature/nio`

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2027,45 +2121,47 @@ private void clientHandshakeSSO(final String db, final OpenIDAuthenticator authe
...
 	}
 
 	public static OAuthFlow getDefaultOAuthFlow() {
 		return Desktop.isDesktopSupported() ? OAuthFlow.AUTHORIZATION_CODE : OAuthFlow.DEVICE_GRANT;
 	}
 
-	private void clientHandshakeOktaNativeSSO(final String db, final OpenIDAuthenticator authenticator, final boolean shouldRequestVersion) throws Exception {
+	private CompletionStage<SQLResult<Void>> clientHandshakeOktaNativeSSO(final String db, final OpenIDAuthenticator authenticator, final boolean shouldRequestVersion) throws Exception {
+
+		CompletionStage<SQLResult<Void>> OktaNativeSSOCompletionStage = SQLResult.voidFuture();
```
jwilliams-ocient, line 2163: (uber uber nit)

lower camel case for local variables. We should probably come up with a style guide for the library...
```suggestion
		CompletionStage<SQLResult<Void>> oktaNativeSSOCompletionStage = SQLResult.voidFuture();
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2549,7 +2648,8 @@ public void connect() throws Exception
 		connect(this.info.getIp(), this.info.getPortNum());
 		try
 		{
-			clientHandshake(user, pwd, database, false);
+			// FIXME Blocking Call
+			blockingGet(clientHandshake(user, pwd, database, false));
```
jwilliams-ocient, line 2652: Prefer `blockingGetSQLResult` for methods that return `CompletionStage<SQLResult<?>>` since `blockingGet(...)` returns an IOException which has an underlying cause of a SQLException. So the catch block would have to inspect the cause of the IOException to ensure an SQLException wasn't marshaled
```suggestion
			blockingGetSQLResult(clientHandshake(user, pwd, database, false)).get();
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3491,7 +3591,8 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
 
 			try
 			{
-				clientHandshake(user, pwd, database, shouldRequestVersion);
+				// FIXME Blocking Call
+				blockingGet(clientHandshake(user, pwd, database, shouldRequestVersion));
```
jwilliams-ocient, line 3595: (same)

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3554,7 +3655,8 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
 
 				try
 				{
-					clientHandshake(user, pwd, database, shouldRequestVersion);
+					// FIXME Blocking Call
+					blockingGet(clientHandshake(user, pwd, database, shouldRequestVersion));
```
jwilliams-ocient, line 3659: (same)

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3606,7 +3708,8 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
...
```
jwilliams-ocient, line 3712: (same)

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3655,7 +3758,8 @@ public void reconnect(final boolean shouldRequestVersion) throws IOException, SQ
 
 		try
 		{
-			clientHandshake(user, pwd, database, shouldRequestVersion);
+			// FIXME Blocking Call
+			blockingGet(clientHandshake(user, pwd, database, shouldRequestVersion));
```
jwilliams-ocient, line 3762: (same)

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3845,7 +3949,8 @@ private void redirectInternal(final String host, final int port, final boolean s
 
 			try
 			{
-				clientHandshake(user, pwd, database, shouldRequestVersion);
+				// FIXME Blocking Call
+				blockingGet(clientHandshake(user, pwd, database, shouldRequestVersion));
```
jwilliams-ocient, line 3953: (same)

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3883,7 +3988,8 @@ private void redirectInternal(final String host, final int port, final boolean s
 
 				try
 				{
-					clientHandshake(user, pwd, database, shouldRequestVersion);
+					// FIXME Blocking Call
+					blockingGet(clientHandshake(user, pwd, database, shouldRequestVersion));
```
jwilliams-ocient, line 3992: (same)

src/main/java/com/ocient/jdbc/XGConnection.java
jwilliams-ocient: EDIT: I actually nerfed this method on your reconnect() branch so no need to change this on this PR actually. 😃

src/main/java/com/ocient/jdbc/XGConnection.java
awang-ocient: Some of the clientHandshake methods still throw Exception for various reasons (catching RuntimeException, etc). Do we need to remove all throws and try/catch blocks in all handshake methods?

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,597 +1559,697 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+
+		// .thenCompose(handshakeSuccessOrEx -> handshakeSuccessOrEx
+		// 		.recoverAsync(ex -> disconnect(false)
+		// 							.thenApply(closeTransportSuccessOrEx -> { 
+
+		// 								if(closeTransportSuccessOrEx.isException())
+		// 									LOGGER.log(Level.WARNING, closeTransportSuccessOrEx.getException(), () -> "Failed to close socket in clientHandshakeGCM");
+
+		// 								return handshakeSuccessOrEx;
+		// 							})));
```
jwilliams-ocient, line 1836: Yea should be safe to remove, I recall adding an assertion that we disconnect exactly once on connect failures.
```suggestion
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,597 +1559,697 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+
+		// .thenCompose(handshakeSuccessOrEx -> handshakeSuccessOrEx
+		// 		.recoverAsync(ex -> disconnect(false)
+		// 							.thenApply(closeTransportSuccessOrEx -> { 
+
+		// 								if(closeTransportSuccessOrEx.isException())
+		// 									LOGGER.log(Level.WARNING, closeTransportSuccessOrEx.getException(), () -> "Failed to close socket in clientHandshakeCBC");
+
+		// 								return handshakeSuccessOrEx;
+		// 							})));
```
jwilliams-ocient, line 2024: (same here)
```suggestion
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,597 +1559,697 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+
+									return handshakeSuccessOrEx;
+								});
+					}));
+		
+		if (shouldRequestVersion && serverVersion.equals(""))
+			handshakeSecurityTokenCompletionStage = handshakeSecurityTokenCompletionStage.thenCompose(successOrEx -> successOrEx.mapSuccessAsync(success -> fetchServerVersion()));
+		
+		return handshakeSecurityTokenCompletionStage;
 	}
```
jwilliams-ocient, line 2246: I _think_ we can get rid of this entire block since `initTransportAndBeginHandshake()` now does the connect error handling (i.e. refresh / disconnect, or try another interface)

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2195,35 +2299,47 @@ private void clientHandshakeSSO(final String db, final OpenIDAuthenticator authe
...
 
 			// Now send the tokens to the database for validation
-			clientHandshakeGCM(username, password, db, shouldRequestVersion, HandshakeType.SSO);
+			SSOHandshakeCompletionStage = clientHandshakeGCM(username, password, db, shouldRequestVersion, HandshakeType.SSO)
+											.thenApply(successOrEx -> {
+												LOGGER.log(Level.INFO, () -> "Handshake SSO without token finished");
+												return successOrEx;
+											});
 		}
 		catch(RuntimeException e)
```
jwilliams-ocient, line 2308: I know you didn't add this, but I think we should remove these catch runtime exception blocks

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2195,35 +2299,47 @@ private void clientHandshakeSSO(final String db, final OpenIDAuthenticator authe
 			}
 
 			// Now send the tokens to the database for validation
-			clientHandshakeGCM(username, password, db, shouldRequestVersion, HandshakeType.SSO);
+			SSOHandshakeCompletionStage = clientHandshakeGCM(username, password, db, shouldRequestVersion, HandshakeType.SSO)
+											.thenApply(successOrEx -> {
+												LOGGER.log(Level.INFO, () -> "Handshake SSO without token finished");
+												return successOrEx;
+											});
```
jwilliams-ocient, line 2306: I think we should remove the error handling and the fetch server version stuff to DRY this up.
```suggestion
			return clientHandshakeGCM(username, password, db, shouldRequestVersion, HandshakeType.SSO);
```

We can move the log message to the common code path in `clientHAndshake()` and include the handshake type in the message.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2273,7 +2389,11 @@ private void clientHandshakeOktaNativeSSO(final String db, final OpenIDAuthentic
 			}
 
 			// Now send the tokens to the database for validation
-			clientHandshakeGCM(username, password, db, shouldRequestVersion, HandshakeType.OKTA_NATIVE_SSO);
+			OktaNativeSSOCompletionStage = clientHandshakeGCM(username, password, db, shouldRequestVersion, HandshakeType.OKTA_NATIVE_SSO)
+											.thenApply(successOrEx -> {
+												LOGGER.log(Level.INFO,() -> "Handshake SSO without token finished");
+												return successOrEx;
+											});
```
jwilliams-ocient, line 2396: Same here
```suggestion
			return clientHandshakeGCM(username, password, db, shouldRequestVersion, HandshakeType.OKTA_NATIVE_SSO);
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2283,27 +2403,33 @@ private void clientHandshakeOktaNativeSSO(final String db, final OpenIDAuthentic
...
+		CompletionStage<SQLResult<Void>> handshakeOktaSessionCompletionStage = SQLResult.voidFuture();
+
 		try
 		{
 			LOGGER.log(Level.INFO, "Beginning Okta Session Token Exchange");
 
-
 			// This handshake can only be used with the JDBC CLI
 			if (!CLI.isInitialized() || CLI.getReader() == null) {
 				throw new SQLFeatureNotSupportedException(String.format("The JDBC CLI is required to use the %s handshake", HandshakeType.OKTA_NATIVE_SSO));
```
jwilliams-ocient, line 2435: If this is the call that requires us to keep the `throws` on the interface, we can give it back to the caller via complete stage
```suggestion
				return SQLResult.failAsync(new SQLFeatureNotSupportedException(String.format("The JDBC CLI is required to use the %s handshake", HandshakeType.OKTA_NATIVE_SSO)));
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2343,11 +2469,11 @@ private void clientHandshakeOktaSessionToken(final String db, final OpenIDAuthen
...
-						clientHandshakeGCM(token.getTokenHint(), token.getToken(), db, shouldRequestVersion, HandshakeType.OKTA_SESSION_TOKEN);
-
-						LOGGER.log(Level.INFO, "Okta Session Token Exchange complete");
-
-						return;
+						handshakeOktaSessionCompletionStage = clientHandshakeGCM(token.getTokenHint(), token.getToken(), db, shouldRequestVersion, HandshakeType.OKTA_SESSION_TOKEN)
+								.thenApply(successOrEx -> { 
+									LOGGER.log(Level.INFO, () -> "Okta Session Token Exchange complete"); 
+									return successOrEx;
+								});
```
jwilliams-ocient, line 2476: This changeset is going to be beautiful, so much delete boilerplate! 😄 
```suggestion
						return clientHandshakeGCM(token.getTokenHint(), token.getToken(), db, shouldRequestVersion, HandshakeType.OKTA_SESSION_TOKEN);
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2343,11 +2469,11 @@ private void clientHandshakeOktaSessionToken(final String db, final OpenIDAuthen
...
 					}
 				case DEVICE_GRANT:
 					throw new SQLFeatureNotSupportedException(String.format("Okta Session Token requires the %s %s",
```
jwilliams-ocient, line 2479:
```suggestion
					return SqlResult.failOrMarshalAsync(new SQLFeatureNotSupportedException(String.format("Okta Session Token requires the %s %s", OAuthFlow.AUTHORIZATION_CODE.key(), ConnectionProperty.SSO_O_AUTH_FLOW.key())));
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -3750,7 +3887,7 @@ private CompletionStage<SQLResult<Void>> initTransportAndBeginHandshake(
 						return SQLResult.failAsync(initSuccessOrEx.getException());
 					}
 
-					return clientHandshakeAsync(user, pwd, database, shouldRequestVersion)
+					return clientHandshake(user, pwd, database, shouldRequestVersion)
```
jwilliams-ocient, line 3890: 😍

jwilliams-ocient (commented):
So close, just some more code we can delete

src/main/java/com/ocient/jdbc/XGConnection.java
jwilliams-ocient: The `fetchServerVErsion()` stuff should also be done within `clientHandshake()` using a `thenCompose()/mapSuccessAsync()` chain on the future that is currently being returned.

Bye bye boilerplate

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,599 +1559,646 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+				if (ccr2.getRedirect())
+				{
+					LOGGER.log(Level.INFO, "Redirect command in ClientConnectionGCM2Response from server");
+					return redirect(ccr2.getRedirectHost(), ccr2.getRedirectPort(), shouldRequestVersion);
+				}
 			}
-			else
-			{
-				hand2.setForce(false);
+			catch(Exception e) {
```
jwilliams-ocient, line 1832: I think I failed to remove the `throws Eception` from `saveSecondaryInterfaces()` so I think you can get rid of this. ERule of thumb is to remove the `throws` from offending methods until the compiler complains that an exception hasn't been caught.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,599 +1559,646 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
 				if (!authenticator.isPresent()) {
-					LOGGER.log(Level.INFO, "No OpenID Authenticators found");
-					throw new SQLException("No OpenID Authenticator(s) found!");
+					LOGGER.log(Level.INFO, "No Okta Authenticators found");
+					return SQLResult.failAsync(new SQLException("No Okta Authenticator(s) found!"));
 				}
 
-				clientHandshakeSSO(db, authenticator.get(), shouldRequestVersion);
-			} else {
+				return clientHandshakeOktaNativeSSO(db, authenticator.get(), shouldRequestVersion);
```
jwilliams-ocient, line 1660:
```suggestion
				return clientHandshakeOktaNativeSSO(db, authenticator.get());
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,599 +1559,646 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+					.filter(a -> a.getIssuer().contains("okta.com"))
+					.findFirst();
+				
+				if (!authenticator.isPresent()) {
+					SQLException e = new SQLException("No Okta Authenticator(s) found!");
+					LOGGER.log(Level.INFO, e, () -> "No Okta Authenticators found");
+					return SQLResult.failAsync(e);
+				}
+				
+				return clientHandshakeOktaSessionToken(db, authenticator.get(), shouldRequestVersion);
```
jwilliams-ocient, line 1683:
```suggestion
				return clientHandshakeOktaSessionToken(db, authenticator.get());
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,599 +1559,646 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+			} else {
+				// GCM
+				return clientHandshakeGCM(userid, pwd, db, shouldRequestVersion, handshake);
```
jwilliams-ocient, line 1686:
```suggestion
				return clientHandshakeGCM(userid, pwd, db, handshake);
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,599 +1559,646 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+		final ClientWireProtocol.Request.Builder b2 = ClientWireProtocol.Request.newBuilder();
+		b2.setType(ClientWireProtocol.Request.RequestType.CLIENT_CONNECTION_GCM2);
+		b2.setClientConnectionGcm2(msg2);
+
+		return b2.build();
 	}
 
 	// DB-15559 make this and clientHandshakeCBC share code. Didn't have time to do this right now.
-	private void clientHandshakeGCM(final String userid, final String pwd, final String db, final boolean shouldRequestVersion, final HandshakeType handshake) throws Exception
+	private CompletionStage<SQLResult<Void>> clientHandshakeGCM(final String userid, final String pwd, final String db, final boolean shouldRequestVersion, final HandshakeType handshake)
```
jwilliams-ocient, line 1746:
```suggestion
	private CompletionStage<SQLResult<Void>> clientHandshakeGCM(final String userid, final String pwd, final String db, final HandshakeType handshake)
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,599 +1559,646 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
+		oneShotForce = false;
+		final ClientConnectionSecurityToken msg = builder.build();
+		ClientWireProtocol.Request.Builder b2 = ClientWireProtocol.Request.newBuilder();
+		b2.setType(ClientWireProtocol.Request.RequestType.CLIENT_CONNECTION_SECURITY_TOKEN);
+		b2.setClientConnectionSecurityToken(msg);
+
+		return b2.build();
+	}
+
+	private CompletionStage<SQLResult<Void>> clientHandshakeSecurityToken(final String db, final boolean shouldRequestVersion)
```
jwilliams-ocient, line 2154:
```suggestion
	private CompletionStage<SQLResult<Void>> clientHandshakeSecurityToken(final String db)
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2191,43 +2242,24 @@ private void clientHandshakeSSO(final String db, final OpenIDAuthenticator authe
...
 
 	public static OAuthFlow getDefaultOAuthFlow() {
 		return Desktop.isDesktopSupported() ? OAuthFlow.AUTHORIZATION_CODE : OAuthFlow.DEVICE_GRANT;
 	}
 
-	private void clientHandshakeOktaNativeSSO(final String db, final OpenIDAuthenticator authenticator, final boolean shouldRequestVersion) throws Exception {
-		try
-		{
-			LOGGER.log(Level.INFO, "Beginning Okta Native SSO Flow");
+	private CompletionStage<SQLResult<Void>> clientHandshakeOktaNativeSSO(final String db, final OpenIDAuthenticator authenticator, final boolean shouldRequestVersion) {
```
jwilliams-ocient, line 2258:
```suggestion
	private CompletionStage<SQLResult<Void>> clientHandshakeOktaNativeSSO(final String db, final OpenIDAuthenticator authenticator) {
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2273,56 +2305,37 @@ private void clientHandshakeOktaNativeSSO(final String db, final OpenIDAuthentic
...
+			return SQLResult.failOrMarshalAsync(e);
 		}
-		LOGGER.log(Level.INFO, "Handshake SSO without token finished");
 	}
 
-	private void clientHandshakeOktaSessionToken(final String db, final OpenIDAuthenticator authenticator, final boolean shouldRequestVersion) throws Exception {
-		try
-		{
-			LOGGER.log(Level.INFO, "Beginning Okta Session Token Exchange");
+	private CompletionStage<SQLResult<Void>> clientHandshakeOktaSessionToken(final String db, final OpenIDAuthenticator authenticator, final boolean shouldRequestVersion) {
```
jwilliams-ocient, line 2314:
```suggestion
	private CompletionStage<SQLResult<Void>> clientHandshakeOktaSessionToken(final String db, final OpenIDAuthenticator authenticator) {
```

awang-ocient:
Still need to propagate `shouldRequestVersion` because there is a circular dependency in the function calls.

Example:

clientHandshake -> clientHandshakeGCM -> redirect -> redirectInternal -> initTransportAndBeginHandshake -> clientHandshake

`shouldRequestVersion` is propagated through all these calls.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,629 +1559,676 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
-				return;
-			}
+		return clientHandshakeExecute(userid, pwd, db, shouldRequestVersion)
+            .thenCompose(handshakeTypeOrEx -> handshakeTypeOrEx
+                .mapSuccessAsync(handshakeType -> {
+                    /**
+                     * On success, retrieve the server's version.
+                     *
+                     * TODO, this is dumb... we should include the server version in the handshake response and save 1RT
+                     */
```
jwilliams-ocient, line 1571: RE not being able to remove `shouldResultVersion`:

We can clean that up if/when we include the server version in the actual handshake response. There are a lot of unnecessary RT's because things were added to the protocol later on, exposing the server version in this case. For instance, we should have a `ResetAllParameters` method so we don't have N RTs where N is the number of parameters.

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,629 +1559,676 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
-			{
-				LOGGER.log(Level.WARNING, String.format("Exception %s occurred during handshake with message %s", e.toString(), e.getMessage()));
-				throw e;
+				crypto = clientHandshakePublicKeyCrypto(ccr.getPubKey());
 			}
 			catch (final Exception e)
 			{
 				LOGGER.log(Level.WARNING, String.format("Exception %s occurred during handshake with message %s", e.toString(), e.getMessage()));
-				throw e;
+				return e;
```
jwilliams-ocient, line 1785: does this compile?

I would expect the return type of `SQLResult<T>#mapSuccess` to be `SQLResult<U>` where `U` is `EncryptedHandshake` in this case.

Maybe this?
```suggestion
				return SQLResult.failOrMarshal(e);
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,629 +1559,676 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
 
-			final byte[] iv = ivString.toByteArray();
 			// We are using a 16 byte authentication tag on the server side. 16 * 8 = 128.
+			final byte[] iv = ivString.toByteArray();
 			final GCMParameterSpec gps = new GCMParameterSpec(128, iv);
+			EncryptedHandshake encryptedHand;
+			try {
+				encryptedHand = clientHandshakeEncryption(crypto, gps, "AES/GCM/NoPadding");
+			} catch (Exception e) {
+				return e;
```
jwilliams-ocient, line 1795:
```suggestion
				return SQLResult.failOrMarshal(e);
```

src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -1555,629 +1559,676 @@ private void saveSecondaryInterfaces(final List<String> newCmdComps, final List<
...
-			final SecretKey aesKey = new SecretKeySpec(key, "AES");
-			final SecretKey hmacKey = new SecretKeySpec(macKey, "AES");
-
-			// Create a Cipher for encrypting the data using the key we created.
-			Cipher encryptCipher;
-
-			encryptCipher = Cipher.getInstance("AES/GCM/NoPadding");
-			// Initialize the Cipher with key and parameters
-			encryptCipher.init(Cipher.ENCRYPT_MODE, aesKey, gps);
+			return encryptedHand;
```
jwilliams-ocient, line 1798: and this?
```suggestion
			return SQLResult.success(encryptedHand);
```

jwilliams-ocient (commented):
One last question about the return type of a few `mapSuccess` blocks

src/main/java/com/ocient/jdbc/XGConnection.java
awang-ocient: Yes it compiles and I ran the unit tests. Looks like `mapSuccess` wraps the return value inside a SQLResult. 

Both `return encryptedHand` and `return e` are inside the lambda function passed into `mapSuccess`

awang-ocient: If we do `return SQLResult.failOrMarshal(e)`, then `mapSuccess` would return type `SQLResult<SQLResult<Object>>` instead of the desired `SQLResult<Object>`

jwilliams-ocient: For instance, this doesn't compile for me:
```
@Test
public void testMapSuccessToException() throws Exception {
    SQLException ex = new SQLException("My exception");

    SQLResult<Void> result = SQLResult
        .voidResult()
        .mapSuccess(o -> {
            return ex;
        });

    assertTrue(result.isException());
    assertFalse(result.isSuccess());
    assertTrue("Expected the unmarshaled exception instance", ex == result.getException());
}
```

awang-ocient: U is Object. I end up casting in the logic below which isn't great. We can switch back to `thenCompose` + `mapSuccessAsync` which allows us to guarantee a return type of `SQLResult<EncryptedHandshake>`?

awang-ocient:
```
.thenCompose(ccrOrEx -> ccrOrEx.mapSuccessAsync(ccr -> {
			// Handle first handshake response
			final ByteString ivString = ccr.getIv();
			ClientHandshakeCrypto crypto;

			try
			{
				crypto = clientHandshakePublicKeyCrypto(ccr.getPubKey());
			}
			catch (final Exception e)
			{
				LOGGER.log(Level.WARNING, String.format("Exception %s occurred during handshake with message %s", e.toString(), e.getMessage()));
				return SQLResult.failOrMarshalAsync(e);
			}

			// We are using a 16 byte authentication tag on the server side. 16 * 8 = 128.
			final byte[] iv = ivString.toByteArray();
			final GCMParameterSpec gps = new GCMParameterSpec(128, iv);
			EncryptedHandshake encryptedHand;
			try {
				encryptedHand = clientHandshakeEncryption(crypto, gps, "AES/GCM/NoPadding");
			} catch (Exception e) {
				return SQLResult.failOrMarshalAsync(e);
			}

			return SQLResult.successAsync(encryptedHand);
		}))
```

jwilliams-ocient: Ohhh I think I missed that part, hmmm...

I'd prefer _not_ to use `Object` if possible. A ton of legacy code does that, making it hard to do compile-time checks, for instance, ensuring that all possible variants of Object are handled, so I've been _trying_ to refactor that code in a type-safe manner. There's a terrible one in `XGStatement#sendAndReceive`.

Is there a common interface you can extract from the set of non-exceptional types that you expect?

awang-ocient: I think the issue here is that `mapSuccess` wraps the return value of the lambda in SQLResult before returning.

Whereas `mapSuccessAsync` simply returns the return value of the lambda

jwilliams-ocient: Oh, I see what you're trying to do... what you _really_ want here is "flatMap". Common in monadic/functional interfaces and something I didn't have a use case for yet so I didn't add this functionality to `*Result`, but this seems like a good use case!

Here is an [example](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#flatMap-java.util.function.Function-) of this pattern in java's `Optional` implementation.

In our case, we want:
```
default <U> SQLResult<U> flatMapSuccess(final Function<? super T, ? extends SQLResult<U>> fn) {
    return fold(fn::apply, SQLResult::fail);
}
```

"flatMapSuccess" to stay consistent with the "success" "fail"/"ex" pattern.

jwilliams-ocient: > Whereas mapSuccessAsync simply returns the return value of the lambda

Right because that function returns a future that returns the result:
```
default <U> CompletionStage<SQLResult<U>> mapSuccessAsync(final Function<? super T, ? extends CompletionStage<SQLResult<U>>> fn) {
```

Okay, it makes sens why you were trying to do that, I would:
1. Extend the `Result`, `IOResult`, "SQLResult`, and `TransportResult` interfaces with this new method, and then add a test case to CompletableFuturesTest.java

PR #345, opened 2023-04-24
[DB-23687] added JDBC benchmarks

sshajith-ocient:
[DB-23687]

[DB-23687]: https://ocient.atlassian.net/browse/DB-23687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/com/ocient/jdbc/XGConnection.java
```diff
@@ -2381,7 +2381,7 @@ private void closeTransport() throws IOException, SQLException {
 	@Override
 	public void close() throws SQLException {
 		LOGGER.log(Level.INFO, String.format("close() called on the connection: %s", this));
-
+		NETTY_OCIENT_WIRE_V1_TRANSPORT_FACTORY.shutdown();
```
jwilliams-ocient, line 2384: I don't think this can be done in this context because `NETTY_OCIENT_WIRE_V1_TRANSPORT_FACTORY` is global and shared amongst all of the sockets created by the application. So this is only safe to do if:
1. you can guarantee _this_ is the only connection used by the process
2. you can reinitialize the thread group if a new connection is created

This requires a bit of synchronization because the lifecycle of an individual connection object differs from the application's lifecycle. The best option is to keep these threads alive while the JVM is running. You can do this by:
1. Initializing the threads when the class is loaded. This is done using a `static` initializer for `NETTY_EVENT_LOOP_GROUP`.
2. Tearing down the threads when the JVM exits. You can register a "shutdown hook" in a static block below the initializer for `NETTY_EVENT_LOOP_GROUP`.

(2) looks like:
```
private static final EventLoopGroup NETTY_EVENT_LOOP_GROUP = Epoll.isAvailable()
                ? new EpollEventLoopGroup()
                : new NioEventLoopGroup();

static {
  // Tear down the event loop group on JVM exit
  Runtime
    .getRuntime()
    .addShutdownHook(new Thread(NETTY_OCIENT_WIRE_V1_TRANSPORT_FACTORY::shutdown));
}
```

jwilliams-ocient: Probably hard to tell from a cursory glance, but `XGConnection#close` actually doesn't close the socket, it actually returns the socket to an object pool. The socket is only closed when the connection is evicted from the pool, see `XGConnectionFactory#destroyObject(...)` => `XGConnection#destroy` => `XGConnection #closeTransport`.

src/test/java/com/ocient/jdbc/JDBCBenchmarks.java
```diff
@@ -0,0 +1,202 @@
...
+        }
+        m_conn = null;
+    }
+
+    public ResultSet executeAndRetrieveUtility(String sqlStatement) {
+        ResultSet rs = null;
+        try {
+            m_stmt = m_conn.createStatement();
+            m_stmt.execute(sqlStatement);
+            rs = m_stmt.getResultSet();
```
jwilliams-ocient, line 150: We also need to consume all elements in the result set.
```suggestion
            rs = m_stmt.getResultSet();
            while (rs.next()) {
              // consume all rows
            }
```

src/test/java/com/ocient/jdbc/JDBCBenchmarks.java
```diff
@@ -0,0 +1,202 @@
...
+    private static String m_url = "";
+
+    private static AtomicInteger teardownCounter = new AtomicInteger();
+
+    @Param({"blocking", "nonblocking"})
+    private static String m_transport;
+    private static Properties m_properties = new Properties();;
+
+    private static Connection m_conn;
+    private static Statement m_stmt;
```
jwilliams-ocient, line 75: static implies global (at the class loader context w.r.t. to the JVM)
```suggestion
    private Connection m_conn;
    private Statement m_stmt;
```

src/test/java/com/ocient/jdbc/JDBCBenchmarks.java
```diff
@@ -0,0 +1,202 @@
...
+                .include(JDBCBenchmarks.class.getSimpleName())
+                .forks(1)
+                .warmupIterations(2)
+                .measurementIterations(5)
+                .build();
+        Runner r = new Runner(opt);
+        r.run();
+    }
+
+    private static String m_url = "";
```
jwilliams-ocient, line 66: I don't think this should be global, it limits our ability to write test cases with injected properties
```suggestion
    private String m_url = "";
```

src/test/java/com/ocient/jdbc/JDBCBenchmarks.java
```diff
@@ -0,0 +1,202 @@
...
+
+    private static AtomicInteger teardownCounter = new AtomicInteger();
```
jwilliams-ocient, line 68:
```suggestion
```

src/test/java/com/ocient/jdbc/JDBCBenchmarks.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @Param({"blocking", "nonblocking"})
+    private static String m_transport;
```
jwilliams-ocient, line 71:
```suggestion
    private String m_transport;
```

jwilliams-ocient:
Forgot to leave a comment on that review, why are all of the variables in the test global?

